PR #13, opened 2019-10-21
Ensure we always return a non-null version (0.2.1 release candidate)

marco-m-pix4d:
- [ ] update release date if needed
- [ ] last tests and issue release

We were already always returning a non-nil version for the put and
check steps. This commit does the same also for the get step.

Why? Returning a nil version was crashing the ATC for a put step (see
https://github.com/concourse/concourse/pull/4442, fixed in Concourse 5.6.0).
We err on the side of caution and ensure (and explicitly tests) we do the
same also fot check and get.

It is highly unlikely that this was impacting anything since the Cogito
documentation never suggested to use a get step, but better safe than sorry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Cogito is now open source.
 - The cogito.yml sample pipeline is fully parametric.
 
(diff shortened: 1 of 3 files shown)


marco-m-pix4d:
@odormond FYI

odormond:
> @odormond FYI

OK. I Completely forgot about that one. :laughing:

PR #29, opened 2021-03-04
Update tooling

marco-m-pix4d:
- doc: update CHANGELOG
- build: correctly propagate the docker tag to the image
    This fixes the bug:
   ```
    This is the Cogito GitHub status resource.
    Tag: <missing value>, commit: 5dbf3c4296, build date: 2021-03-04
    ```
   To:
   ```
    This is the Cogito GitHub status resource.
    Tag: update-tooling, commit: 5dbf3c4296, build date: 2021-03-04
   ```
- go: update all dependencies
- build: always run tests (do not cache them)
- pipelines/cogito: do not require SSH key, it is a public repo
- build: use gopass for secrets
- Update to Go 1.16
- build: update Task version

Part of: PCI-1767

test pipeline: https://builder.ci.pix4d.com/teams/developers/pipelines/cogito-update-tooling


odormond (approved):

pipelines/cogito.yml
```diff
@@ -57,9 +57,10 @@ resources:
 - name: repo.git
   type: git
   source:
-    uri: git@github.com:((github-owner))/((repo-name)).git
+    uri: https://github.com/((github-owner))/((repo-name)).git
+    #uri: git@github.com:((github-owner))/((repo-name)).git
```
manuel-pix4d, line 61: leftover

CONTRIBUTING.md
```diff
@@ -12,8 +12,8 @@ I care about code quality, readability and tests, so please follow the current s
 
 ### Required
 
-* golang, version >= 1.13
-* docker, version >= 19
+* Go, version >= 1.16
```
manuel-pix4d, line 15: We currently don't have any tracking of version upgrade for cogito. Do you think it makes sense to think of one ? I don't have in mind but might be needed.

pipelines/cogito.yml
manuel-pix4d: Then, shall we remove the private_key above ?

manuel-pix4d (approved):
LGTM. few comments here and there but nothing blocking.

CONTRIBUTING.md
marco-m-pix4d: I think it is OK as-is for the moment.
I plan to move the CI from Travis to GitHub Actions; when that is done, we can reconsider.

pipelines/cogito.yml
marco-m-pix4d: I understand that it can be confusing at first sight.
I left the commented-out line and the SSH private_key there on-purpose: it is up to the owner of the pipeline to adapt according to wether the target repo is public or private.
Your comment makes me think I can make it more evident; I will push a change.

marco-m-pix4d: Changed.

Issue #24, opened 2021-02-14
Add pipeline name as prefix to context


tgolsson:
I'm trying out concourse and setting up for a monorepo/multi-project workflow, and it's easy to end up having test/validate/build jobs in multiple pipelines for the same repository, etc. I'm wondering what your stance is on prefixing the pipeline name to the context, if the information is available? As an example, instead of just `test` for the `test` job it'd be for example `publish-docker-image/test` if this test is part of the publishing pipeline.

marco-m-pix4d:
Hello @tgolsson,
we were using as GH status context the format `PIPELINENAME/JOBNAME`, but this was making impossible our workflow of using branch protection rules, see #11.
On the other hand, I can see the usefulness of prepending the pipeline name.

I would suggest to think how you would use the branch protection rules in your case and, if it would work, I am open to a PR. We would have to introduce this feature as an optional attribute in the source configuration.

tgolsson:
I see the problem! I've not thought too much about branch protection rules and interaction with concourse. From the description I take it you run a multi-repo setup with per-branch pipelines? I've not gotten that deep yet but it seems like the branching story is more of a concourse issue. For me, I'd have the problem that I'd not be able to see all my "test" statuses so I also can't add branch protection rules.

I'm wondering whether maybe this should be a prefix instead, i.e, in the resource declaration we could declare a `status_context_prefix: feature-branch`? That decouples it from branch-based pipelines while still preventing collisions.

marco-m-pix4d:
Currently Concourse doesn't have a out-of-the-box solution for what I call "feature branch pipelines" vs "main branch pipeline".
Forget for a moment wether you have a monorepo or not, this is for the most part independent.
Think about a repo with one project, to make things simple to reason about.

Everything works fine if you use only one pipeline, to represent the status of the main branch, and maybe one pipeline (or job) to represent pull requests.

What we didn't like about that approach is that it goes back to Jenkins: each build in the history of the PR pipeline represents a different, almost random PR. We wanted instead a 1:1 mapping between a feature branch and a pipeline, and that pipeline must be identical (modulo maybe deploy steps) to the main pipeline.

So for example for project banana, branch feat-1 takes pipeline name banana-feat-1 and branch feat-2 takes pipeline name banana-feat-2.

[...]

PR #25, opened 2021-02-23
add context namespace

tgolsson:
Opening a PR for the final version of #24.

This lets a user specify a "namespace" for their contexts, allowing reuse of job-names (contexts) across multiple pipelines. 

With the resource set up like so:

```
# This is the status object that will be updated in each job
- name: gh-status
  type: cogito
  check_every: 1h
  source:
    owner: tgolsson
    repo: concourse-ci-test
    access_token: ((status-token))
    context_namespace: phatik-frontend
```

The final result with how I've configured this in my "test repo" is like this:

image

I've noticed that [telia-oss/github-pr-resource](https://github.com/telia-oss/github-pr-resource) calls the same concept `base_context` which makes sense but is less clear (to me). I'm happy to change it if you'd prefer to be consistent with others.

Issue #23, opened 2020-12-10
Provide a way to override BUILD_JOB_NAME === context on resources.


eitah:
Right now the resource requires that you put all tasks that affect the same job status together in the same job within a pipeline because of the choice to make github "context" ===  BUILD_JOB_NAME. For reference, I believe this is where we set it. https://github.com/Pix4D/cogito/blob/master/resource/resource.go#L213-L215

I'd like to propose allowing users to pass in an optional parameter "context" that overrides the job name  as the CONTEXT arg if not equal to "". Any objections or "gotchas" you can think of? Any reason you'd rather not have this property be configurable? 

I'm happy to PR this if you think the feature has a chance.

marco-m-pix4d:
Hello @eitah,
I appreciate both that you took the time to look into the code and that you opened this ticket instead of jumping to a PR.
I am trying to understand better the use case. I see what you want to do, I would like to understand the why.

To make a concrete example, I think that you would like to be able to have a pipeline with, say, 2 jobs as follows:

* `jobA`, containing among the others, task `tA`.
* `jobB`, containing among the others, task `tB`.

Where for some reasons tasks `tA` and `tB` are related, and you would like that `jobB` set the GH status with a `context` that is different from the name `jobB`.

The question I have is this: if you already have the two jobs, it means that they have a certain logical grouping. So, you could push GH status two times, the first time on completion of `jobA`, and the second time on completion of `jobB`. If the names `jobA` and `jobB` are meaningful, they work well as GH context.

For example, if first job is `build` and second job is `test`, the same commit would be decorated with contexts `build` and `test`. This would also give more granularity.

What am I missing?

eitah:
I really like and agree with that point about mandating each job pipeline close its own statuses as a way to enforce granularity. I'll pitch that to the Dev that first raised the bug. Without hearing from him I'd assume his objection might be:

- More YAML to configure the job closure step per pipeline. 
- The code change to enable an override of the job name === context is relatively slight (1-3 lines?) which is nothing in go terms.

Thanks!
Eli

eitah:
He's good with just having each job manage its own statuses so I think this feature is not required. No code == good code :) thanks again!

eitah:
Update after a week of closing statuses per pipeline:

> i really don’t like that cogito resource =\

>they make it so complicated to just 1) when the pipeline starts, mark a commit/PR pending; 2) anywhere along the way, if something fails, mark it failed; 3) at the very end, mark it a success

[...]

PR #33, opened 2021-03-08
Add context_prefix optional decoration

marco-m-pix4d:
In the past, we used to decorate the "status context" as `PIPELINE_NAME/JOB_NAME`, but we had to remove the prefix (`PIPELINE_NAME`) because we could not use the GitHub branch protection rules (see #11).

This PR brings back the context prefix (optional), for workflows where only the main pipeline is used. See [Effects on GitHub](https://github.com/Pix4D/cogito/blob/add-context-prefix/README.md#effects-on-github) of the README for a visual explanation:

gh-ui-decorated

Thanks @tgolsson for the discussion (#24) and for the initial implementation (#25). Thanks @eitah for a similar discussion (#23).

@tgolsson: I wanted to add tests for this feature, but I needed to do some refactoring before being able to do so, so I went ahead and did also the implementation. Your contribution is recognized as the author of b9f477c in this branch.

Better reviewed commit-per-commit.

Closes: #24, #25.


resource/resource_test.go
```diff
@@ -134,53 +126,63 @@ func TestOut(t *testing.T) {
...
+					"owner":          cfg.Owner,
+					"repo":           cfg.Repo,
+					"context_prefix": "cocco"},
+				defParams,
+				defEnv,
+			},
+			want{
+				defVersion,
+				defMeta,
+				map[string]string{"context": "cocco/" + defEnv.Get("BUILD_JOB_NAME")},
```
marco-m-pix4d, line 183: This is what we expect from the test. Since the HTTP payload is JSON encoded, we don't want to pass the full JSON object, we want to pass only what we actually want to test. The Go json package allows to use a map, which is perfect for this use (see also comments below)

resource/resource_test.go
```diff
@@ -134,53 +126,63 @@ func TestOut(t *testing.T) {
...
-			want{defVersion, defMeta, nil},
+			want{defVersion, defMeta, nil, nil},
+		},
+		{"source: optional: context_prefix",
+			in{
+				oc.Source{
+					"access_token":   cfg.Token,
+					"owner":          cfg.Owner,
+					"repo":           cfg.Repo,
+					"context_prefix": "cocco"},
```
marco-m-pix4d, line 176: we set the optional context_prefix here

resource/resource_test.go
```diff
@@ -189,6 +191,31 @@ func TestOut(t *testing.T) {
 			inDir, teardown := setup(t, defDir, sshRemote(cfg.Owner, cfg.Repo), cfg.SHA, cfg.SHA)
 			defer teardown(t)
 
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
marco-m-pix4d, line 194: The Go way of mocking an HTTP server

resource/resource_test.go
```diff
@@ -189,6 +191,31 @@ func TestOut(t *testing.T) {
...
+				w.WriteHeader(http.StatusCreated)
+				fmt.Fprintln(w, "Anything goes...")
+
+				if tc.want.body != nil {
+					buf, _ := io.ReadAll(r.Body)
+					var bm map[string]string
+					if err := json.Unmarshal(buf, &bm); err != nil {
+						t.Fatalf("parsing JSON body: %v", err)
+					}
+					for k, v := range tc.want.body {
```
marco-m-pix4d, line 204: if the test case passes us a non-nil map, we decode the JSON object into the `bm` map and then validate that each k,v pair in `tc.want.body` (in this case: `map[string]string{"context": "cocco/" + defEnv.Get("BUILD_JOB_NAME")})` is also present in the JSON.

resource/resource_test.go
```diff
@@ -189,6 +191,31 @@ func TestOut(t *testing.T) {
...
+						if bm[k] != v {
+							t.Errorf("\nbody[%q]: got: %q; want: %q", k, bm[k], v)
+						}
+					}
+				}
+
+			}))
+			savedAPI := github.API
+			github.API = ts.URL
+			defer func() {
```
marco-m-pix4d, line 214: on exit from the subtest, tear down the mock server and especially remove the mock override from `github.API`, otherwise it will break the subsequent tests.

marco-m-pix4d:
@tgolsson this will become release 0.5.0. I would appreciate if you could test the Docker tag `add-context-prefix`, thanks.

odormond (approved):

tgolsson:
@marco-m-pix4d Thank you, very nice! I'll try to test this ASAP, but it'll likely have to wait until the weekend.

aliculPix4D (approved):

PR #34, opened 2021-03-08
add the optional put parameter `context`

### LINKED ISSUE #23: Provide a way to override BUILD_JOB_NAME === context on resources.

marco-m-pix4d:
add the optional put parameter `context`, that will override the default current job name.

See the README for more information.

Closes #23

### DIFF
--- a/resource/resource_test.go
+++ b/resource/resource_test.go
@@ -184,6 +184,24 @@ func TestOut(t *testing.T) {
 				nil,
 			},
 		},
+		{"put step: default context",
+			in{defSource, defParams, defEnv},
+			want{
+				defVersion,
+				defMeta,
+				map[string]string{"context": defEnv.Get("BUILD_JOB_NAME")},
+				nil,
+			},
+		},
+		{"put step: optional: context",
+			in{defSource, oc.Params{"state": "error", "context": "bello"}, defEnv},
+			want{
+				defVersion,
+				defMeta,
+				map[string]string{"context": "bello"},
+				nil,
+			},
+		},
 	}
 
 	for _, tc := range testCases {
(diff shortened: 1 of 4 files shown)


resource/resource_test.go
```diff
@@ -184,6 +184,24 @@ func TestOut(t *testing.T) {
 				nil,
 			},
 		},
+		{"put step: default context",
+			in{defSource, defParams, defEnv},
+			want{
+				defVersion,
+				defMeta,
+				map[string]string{"context": defEnv.Get("BUILD_JOB_NAME")},
```
marco-m-pix4d, line 192: by default we want to see the job name

resource/resource_test.go
```diff
@@ -184,6 +184,24 @@ func TestOut(t *testing.T) {
...
+				nil,
+			},
+		},
+		{"put step: optional: context",
+			in{defSource, oc.Params{"state": "error", "context": "bello"}, defEnv},
+			want{
+				defVersion,
+				defMeta,
+				map[string]string{"context": "bello"},
```
marco-m-pix4d, line 201: if on the other hand we override, we want to see the override

odormond (approved):

aliculPix4D (approved):

PR #37, opened 2021-07-02
Pci 1937 better error msg

aliculPix4D:
CODE before this PR:
```
  left := []string{"github.com", owner, repo} 
  right := []string{gu.Host, gu.Owner, gu.Repo}
  for i, l := range left {
    r := right[i]
    if strings.ToLower(l) != strings.ToLower(r) {
      return fmt.Errorf("remote: %v: got: %q; want: %q: %w", remote, r, l, errWrongRemote)
    }
  }
```

- left is a given configuration set under source configuration
- right is the data parsed from `.git/config`
- error is raised if they don't match

By looking at the error in the ticket
`remote: git@github.com:pix4d/fly_helper.git: got: "fly_helper"; want: "fly_helper.git": wrong git remote`
We can identify the following:
- remote is the value parsed from the `.git/config`, while string `"fly_helper"` just after the got is obtained by parsing the remote string.
- on the other hand string `"fly_helper.git"` arrives from the cogito resource source configuration, which means that `"fly_helper.git"` is the wrong value.
- if we continue from here the http request will fail because `https://github.com/pix4d/fly_helper.git` does not exists while `https://github.com/pix4d/fly_helper` does
- this points us that we put the wrong value in our pipeline resource configuration. 

```
WRONG:
  - name: gh-status
    type: cogito
    check_every: 24h
    source:
      owner: pix4d
      repo: fly_helper.git
      access_token: ((github_repo_status_token))

```
Correct behavior is to stop the code execution here, which is exactly the current behavior. But we should print a better error message i.e like: "cogito source: you have set the wrong repo name"

I additionally added two boy-scout commits. They are a part of this PR because otherwise I could not run the e2e tests.
Note that: https://github.com/Pix4D/cogito-test-read-only do not exist.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -45,10 +45,6 @@ tasks:
         sh: gopass show cogito/test_commit_sha
       COGITO_TEST_OAUTH_TOKEN:
         sh: gopass show cogito/test_oauth_token
-      COGITO_TEST_READ_ONLY_COMMIT_SHA:
-        sh: gopass show cogito/test_read_only_commit_sha
-      COGITO_TEST_READ_ONLY_REPO_NAME:
-        sh: gopass show cogito/test_read_only_repo_name
       COGITO_TEST_REPO_NAME:
         sh: gopass show cogito/test_repo_name
       COGITO_TEST_REPO_OWNER:
--- a/resource/resource.go
+++ b/resource/resource.go
@@ -335,7 +335,7 @@ func repodirMatches(dir, owner, repo string) error {
 	right := []string{gu.Host, gu.Owner, gu.Repo}
 	for i, l := range left {
 		r := right[i]
-		if strings.ToLower(l) != strings.ToLower(r) {
+		if !strings.EqualFold(l, r) {
 			return fmt.Errorf("remote: %v: got: %q; want: %q: %w", remote, r, l, errWrongRemote)
 		}
 	}
--- a/resource/resource.go
+++ b/resource/resource.go
@@ -336,7 +336,9 @@ func repodirMatches(dir, owner, repo string) error {
 	for i, l := range left {
 		r := right[i]
 		if !strings.EqualFold(l, r) {
-			return fmt.Errorf("remote: %v: got: %q; want: %q: %w", remote, r, l, errWrongRemote)
+			return fmt.Errorf("resource source configuration and git repository are incompatible.\nGit remote: %q\n"+
+				"Resource config: host: github.com, owner: %q, repo: %q. %w", remote, owner, repo, errWrongRemote,
+			)
 		}
 	}
 	return nil
--- a/resource/resource_test.go
+++ b/resource/resource_test.go
@@ -348,7 +348,7 @@ func TestRepoDirMatches(t *testing.T) {
 		{"dir is not a repo", "not-a-repo", "dummyurl", os.ErrNotExist},
 		{"bad .git/config", "repo-bad-git-config", "dummyurl", errKeyNotFound},
 		{"repo with wrong HTTPS remote", "a-repo", httpsRemote("owner", "repo"), errWrongRemote},
-		{"repo with wrong SSH remote", "a-repo", sshRemote("owner", "repo"), errWrongRemote},
+		{"repo with wrong SSH remote or wrong source config", "a-repo", sshRemote("owner", "repo"), errWrongRemote},
 		{"repo with good SSH remote", "a-repo", sshRemote(wantOwner, wantRepo), nil},
 		{"repo with good HTTPS remote", "a-repo", httpsRemote(wantOwner, wantRepo), nil},
 	}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,10 +46,9 @@ We require environment variables (as opposed to using a configuration file) to p
 * To enable any contributor to run their own tests without having to edit any file.
 * To securely store secrets!
 
-## Default test repositories
+## Default test repository
 
 * https://github.com/Pix4D/cogito-test-read-write
-* https://github.com/Pix4D/cogito-test-read-only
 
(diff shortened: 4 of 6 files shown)


Taskfile.yml
```diff
@@ -45,10 +45,6 @@ tasks:
         sh: gopass show cogito/test_commit_sha
       COGITO_TEST_OAUTH_TOKEN:
         sh: gopass show cogito/test_oauth_token
-      COGITO_TEST_READ_ONLY_COMMIT_SHA:
```
marco-m-pix4d, line 48: so there are no tests using these variables?

resource/resource.go
```diff
@@ -335,8 +335,10 @@ func repodirMatches(dir, owner, repo string) error {
 	right := []string{gu.Host, gu.Owner, gu.Repo}
 	for i, l := range left {
 		r := right[i]
-		if strings.ToLower(l) != strings.ToLower(r) {
-			return fmt.Errorf("remote: %v: got: %q; want: %q: %w", remote, r, l, errWrongRemote)
+		if !strings.EqualFold(l, r) {
```
marco-m-pix4d, line 338: could you please add a test ?

Taskfile.yml
aliculPix4D: Not as far I am aware.
Here the list does not include them:
https://github.com/Pix4D/cogito/blob/04e9eb5dd4689591010f91488a739c5b3e14a6ba/help/testhelper.go#L160
and quick search in the repo does not find them anywhere except in this file.
Also, all e2e tests are passing without them.

resource/resource.go
aliculPix4D: I thought this was already covered here:
https://github.com/Pix4D/cogito/blob/04b201ef9f445fb53ec8e551d5a87f130a266a1f/resource/resource_test.go#L351
and it seems that chaining `strings.ToLower(l) != strings.ToLower(r)` to `!strings.EqualFold(l, r)` works fine,
but the problem is my new error message, which corresponds only to the last check (repo name), but not the first two (github url, user/organization)

aliculPix4D: I have "added" an extra test and corrected the code, but I don't like how I did it. Can you give me an advice what should I do here? 
I mean given what you wrote in the second part of the ticket: https://pix4dbug.atlassian.net/browse/PCI-1937 you obviously tough that the error message was not clear enough here, so I want to improve it, but my original comment actually applies only when comparing the last item (repo name)

resource/resource_test.go
```diff
@@ -359,6 +360,9 @@ func TestRepoDirMatches(t *testing.T) {
 
 		t.Run(tc.name, func(t *testing.T) {
 			err := repodirMatches(filepath.Join(inDir, tc.dir), wantOwner, wantRepo)
+			if err != nil {
```
aliculPix4D, line 363: this was pushed by mistake. I will remove it

resource/resource.go
```diff
@@ -335,7 +335,11 @@ func repodirMatches(dir, owner, repo string) error {
...
 	for i, l := range left {
 		r := right[i]
-		if strings.ToLower(l) != strings.ToLower(r) {
+		if !strings.EqualFold(l, r) && i == 2 {
+			return fmt.Errorf("repository name you have set in the source configuration (source:repo) %q doesn't correspond"+
+				" to the parsed git configuration %q. Please correct the repo key value. %w", l, r, errWrongRemote,
+			)
+		} else if !strings.EqualFold(l, r) && i < 2 {
 			return fmt.Errorf("remote: %v: got: %q; want: %q: %w", remote, r, l, errWrongRemote)
 		}
```
odormond, line 344: I would not bother with distinguishing the fields. The important thing is the user has enough info to understand what's wrong. So I would just do:
```suggestion
		if !strings.EqualFold(l, r) {
			return fmt.Errorf("Incompatible git remote and resource configuration.\nGit remote: %q\n"+
				"Resource config: host: github.com, owner: %q, repo: %q", remote, owner, repo,
			)
		}
```

odormond (approved):

resource/resource.go
marco-m-pix4d: > I thought this was already covered here:
You are right, I missed that.

Taskfile.yml
marco-m-pix4d: ok

resource/resource.go
marco-m-pix4d: The key to decide what to print here is to understand who is the audience:

Myself, the author of Cogito, when I wrote the first version of the message thought it was clear.
Then I stumbled upon it when creating the fly_helper pipeline and could not understand it fully! :-)

So I think it is worthwhile to give full information here and carefully select the words we use, for example "git remote" is confusing for the end user in this context.

resource/resource.go
marco-m-pix4d: In my understanding, since there was already a test, no need to add another one.
If you still have questions, please ping me here.

marco-m-pix4d (approved):

resource/resource.go
aliculPix4D: Yes, at the end I did not add any test. What I did for the test is changed the test name to point out that both can be correct (wrong SSH config but it can also be that the source configuration is wrong).

```
-  {"repo with wrong SSH remote", "a-repo", sshRemote("owner", "repo"), errWrongRemote},
+ {"repo with wrong SSH remote or wrong source config", "a-repo", sshRemote("owner", "repo"), errWrongRemote},
```

marco-m-pix4d: ok

resource/resource.go
aliculPix4D: Is this better? `resource source configuration and git repository are incompatible`
by resource source configuration I mean: 
 - what is setup in source key when declaring a resource in pipeline configuration
by git repository I mean:
- local fetched git repository

PR #38, opened 2021-07-02
PCI-1937 Improve error message when token does not have write access to the repo.

aliculPix4D:
- Improve error message when token does not have write access to the repo.

Note:
Github will return `404` error if:
- if repo does not exists
- if user has token with correct scope but no access to the repository i.e `pix4d-service-account-cloud`
- if user has token with wrong scope but no access to the repository i.e `pix4d-service-account-cloud`
- if user has wrong token scope but has at least read access to the repo i.e current situation with `Pix4D-Janus`. 

To actually get a `200` status code in `CanReadRepo()` function when doing `API: GET /repos/:owner/:repo` http request we MUST extend the scope for the token from `repo:status` to `repo`. We don't want to do that. 

Additionally, if we actually reached line 137 we are almost 100% sure that the repo exists. Of course there is a small possibility that someone deleted the repo between Concourse fetched it and we reach this point in the code.

### DIFF
--- a/github/status.go
+++ b/github/status.go
@@ -7,7 +7,6 @@ package github
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


marco-m-pix4d:
For my understanding:

> To actually get a 200 status code in CanReadRepo() function when doing API: GET /repos/:owner/:repo http request we MUST extend the scope for the token from repo:status to repo. We don't want to do that.

This is what you were mentioning the reaper does ?

github/status.go
```diff
@@ -135,61 +134,10 @@ func (s status) Add(sha, state, targetURL, description string) error {
 		// Happy path
 		return nil
 	case http.StatusNotFound:
-		msg := disambiguateError(s)
+		msg := "The user with this token doesn't have write access to the repo or the repo doesn't exists (less likely)"
```
aliculPix4D, line 137: Doing an http request `API: GET /repos/:owner/:repo` and expecting to get status code 200 requires different token scope than `repo:status`. Hence there is no point to keep all of this code logic if we will never change the token scope.
Additionally, even if we change the scope to `repo`, if a does not have at least a read access to the repo we still get 404 status code (i.e with `pix4d-service-account-cloud` account) and we can still not distinguish which case we have.

marco-m-pix4d: I am lost here. I forgot the details, but the code was able to disambiguate. Are you saying that in the tests I was using a token with broader scope ?

aliculPix4D:
> For my understanding:
> 
> > To actually get a 200 status code in CanReadRepo() function when doing API: GET /repos/:owner/:repo http request we MUST extend the scope for the token from repo:status to repo. We don't want to do that.
> 
> This is what you were mentioning the reaper does ?

The token that is used in reaper has full repo scope: 
https://github.com/Pix4D/terraform-tomato/blob/ccb5b66381d30569cef307e4135b155d71af14be/ci/pipelines/reaper-template.yml#L75
but if there is any error while doing this request we exit with:
https://github.com/Pix4D/terraform-tomato/blob/ccb5b66381d30569cef307e4135b155d71af14be/ci/reaper/tasks/reaper/checkbranch.py#L30

github/status.go
aliculPix4D: > Are you saying that in the tests I was using a token with broader scope ?

yes.

> I am lost here. I forgot the details, but the code was able to disambiguate. 

It is if you change the token scope.

aliculPix4D: Note that in my test repo https://github.com/Pix4D/test-cogito I am the admin, but my token has only `repo:status` scope and I get 404 here.

aliculPix4D: Quick check:
`curl -X GET -H "Accept: application/vnd.github.v3+json" -H "Authorization: token here_goes_the_token" https://api.github.com/repos/pix4d/your_repo_name`

and try changing the scope.

odormond: As three things could be going wrong here, I would word the message differently:
```suggestion
		msg := "Either the user and/or its token don't have the right permission on the repo or the repo doesn't exists."+
			" (User must have write access on this repo and token must have scope repo:status)"
```

odormond (approved):

github/status.go
marco-m-pix4d: > Note that the check performed here was very biased towards Pix4D-Janus which has a read access by default and it probably had a token before with full scope.

Ah, I think that's it. It could explain this, good catch.

marco-m-pix4d: I propose the following as message, which conveys all possible cases in a format that is clear for the end user:

```
One of the following happened:

1. The repo <REPO> doesn't exist.
2. The user who issued the token doesn't have write access to the repo.
3. The token doesn't have scope repo:status.
```

marco-m-pix4d (approved):

github/status.go
```diff
@@ -135,61 +134,11 @@ func (s status) Add(sha, state, targetURL, description string) error {
 		// Happy path
 		return nil
 	case http.StatusNotFound:
```
Esysc, line 136: To shorter it why not  something like that:
```
msg := ""
if http.StatusNotFound:
    msg := "Either the user and/or its token don't have the right permission or the repo doesn't exists." +
                " (User must have write access on this repo and token must have scope repo:status)" 
return &StatusError{req.Method + " " + url + msg + OAuthInfo, resp.StatusCode, string(respBody)}   
```

Esysc (approved):

PR #43, opened 2021-08-23
refactoring in preparation of instanced pipelines

marco-m-pix4d:
- better log messages for Info and Debug levels
- doc: more details on integration tests
- doc: how to test instanced pipelines

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ I care about code quality, readability and tests, so please follow the current s
 
 ## Optional
 
-* [gopass] to securely store secrets for end-to-end tests.
+* [gopass] to securely store secrets for integration tests.
 * [gotestsum], for more human-friendly test output. If found in `$PATH`, it will be used in place of `go test`.
 
 # Using Task (replacement of make)
@@ -37,35 +37,23 @@ Run the default tests:
 $ task test
 ```
 
-# The end-to-end tests
+# Integration tests
 
-The end-to-end tests (tests that interact with GitHub) are disabled by default because they require some out of band setup, explained below.
+there are two types of integration tests:
+
+* tests against GitHub
+* tests the Docker image as resource inside Concourse
+
+# Integration tests against GitHub
+
+The integration tests (tests that interact with GitHub) are disabled by default because they require some out of band setup, explained below.
 
 We require environment variables (as opposed to using a configuration file) to pass test configuration. The reason is twofold:
 
 * To enable any contributor to run their own tests without having to edit any file.
 * To securely store secrets!
 
-Run all the tests:
-```
-$ task test-e2e
-```
-
-## Running a specific end-to-end test
-
-Use the `test:env` task target, that runs a shell with available all the secrets needed for the e2e tests.
-
-Run all the subtests of a table-driven test:
-
-```
-$ task test:env -- go test ./github -count=1 -run 'TestUnderstandGitHubStatusFailures'
-```
-
-Run an individual subtest of a table-driven test:
-
-```
-$ task test:env -- go test ./github -count=1 -run 'TestUnderstandGitHubStatusFailures/non_existing_SHA:_Unprocessable_Entity'
-```
+The following sections contain first instructions for the setup, then instructions on how to run the tests.
 
 ## Default test repository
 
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -37,35 +37,23 @@ Run the default tests:
 $ task test
 ```
 
-# The end-to-end tests
+# Integration tests
 
-The end-to-end tests (tests that interact with GitHub) are disabled by default because they require some out of band setup, explained below.
+there are two types of integration tests:
```
odormond, line 42:
```suggestion
There are two types of integration tests:
```

CONTRIBUTING.md
```diff
@@ -162,20 +150,36 @@ cogito/
...
-## Running the end-to-end tests
+## Running the integration tests
 
-We are finally ready to run also the end-to-end tests:
+We are finally ready to run also the integration tests:
 
-```console
-$ task test
+```
+$ task test-e2e
```
odormond, line 158: Wouldn't it be more natural to call them `test-integration`? The doc is always talking of "integration tests" and suddenly the command to run them refers to end-to-end tests.

CONTRIBUTING.md
```diff
@@ -162,20 +150,36 @@ cogito/
...
-The end-to-end tests have the following logic:
+The integration tests have the following logic:
 
 * If none of the environment variables are set, we skip the test.
 * If all of the environment variables are set, we run the test.
 * If some of the environment variables are set and some not, we fail the test. We do this on purpose to signal to the user that the environment variables are misconfigured.
 
+## Running a specific end-to-end test
+
+Use the `test:env` task target, that runs a shell with available all the secrets needed for the e2e tests.
```
odormond, line 169: Here they are referred to as "end-to-end"...

I don't really care if we call them integration or end-to-end. Of course, `e2e` is a tempting name as it's much shorter. I would just like that we settle on a single naming.

resource/resource.go
```diff
@@ -230,24 +235,46 @@ func (r *Resource) Out(
...
+	return bld.String()
+}
+
+func stringify2(xs map[string]string) string {
+	var bld strings.Builder
+	for k, v := range xs {
+		fmt.Fprintf(&bld, "  %s: %v\n", k, v)
+	}
+	return bld.String()
+}
```
odormond, line 276: For my own education, isn't a `string` compatible with `interface{}`? I'm expecting any type to be compatible with it so `stringify1` should be enough in all cases, shouldn't it?

odormond (approved):

resource/resource.go
marco-m-pix4d: This is what I was expecting also :-(
I have to admit I don't have a full explanation.

CONTRIBUTING.md
marco-m-pix4d: You are right. Historically, I was calling them e2e, but that name is somehow misleading, so I renamed to integration, but I missed some. Will fix.

CONTRIBUTING.md
marco-m-pix4d: done

CONTRIBUTING.md
marco-m-pix4d: done

resource/resource.go
marco-m-pix4d: Researching a bit, `string` is compatible with `interface{}`, in the sense that this would work:

```go
	var s string

	func foo (x interface{}) string {
		return fmt.Sprintf("%v", x)
	}

	foo(s)
```

The problem here is that type conversion happens only for "flat" types, and `map[string]string` is not flat.

I could do:

```go
func stringify(x interface{}) string {
	if xs, ok := x.(map[string]interface{}); ok {
		var bld strings.Builder
		for k, v := range xs {
			fmt.Fprintf(&bld, "  %s: %v\n", k, v)
		}
		return bld.String()
	}
	if xs, ok := x.(map[string]string); ok {
		var bld strings.Builder
		for k, v := range xs {
			fmt.Fprintf(&bld, "  %s: %v\n", k, v)
		}
		return bld.String()
	}
	return fmt.Sprintf("  %v\n", x)
}
```

but, besides just being silly, it is even worse, because it would accept at compile time also types that are not maps, while in this case we know it is either `map[string]string` or `map[string]interface{}`.

I might be missing some Go type foo, but this is the less worse I could find.

CONTRIBUTING.md
marco-m-pix4d: done

CONTRIBUTING.md
marco-m-pix4d: done

resource/resource.go
odormond: The nested type conversion make sense. Let's wait for the generic to have a clean version. Apparently they missed the 1.17 release...

marco-m-pix4d: yes, this is one of the few cases where generics would have helped.

aliculPix4D (approved):

PR #26, opened 2021-02-25
support instance vars

lsjostro:
Instance vars is a new feature in concourse 7.0 which we should support. This PR also include a small fix when using a git cache proxy in front of github not accessed over https.

PR #44, opened 2021-08-23
rework of #26: support instanced pipelines

marco-m-pix4d:
This is based on the initial work in #26 by @lsjostro.
Since I could not push to the branch in the fork, I created this new PR.
The authorship of the initial PR is maintained, see commit e254f91.

Instructions to the reviewers:
- review #43 before
- best reviewed commit-per-commit

@aoldershaw I would appreciate if you could have a look :-)


resource/resource_test.go
```diff
@@ -309,6 +309,62 @@ func TestOutE2E(t *testing.T) {
...
+	}{
+		"all defaults": {
+			want: "https://ci.example.com/teams/devs/pipelines/magritte/jobs/paint/builds/42",
+		},
+		"instanced vars 1": {
+			instanceVars: `{"branch":"stable"}`,
+			want:         "https://ci.example.com/teams/devs/pipelines/magritte/jobs/paint/builds/42?vars=%7B%22branch%22%3A%22stable%22%7D",
+		},
+		"instanced vars 2": {
+			instanceVars: `{"branch":"stable","foo":"bar"}`,
```
marco-m-pix4d, line 330: @aoldershaw according to your comment https://github.com/Pix4D/cogito/pull/26#discussion_r598955336 in the previous PR I have the impression I should nest the JSON ?

aoldershaw: Sorry, not 100% sure what you mean by nest here, but AFAICT the implementation you have is correct

odormond (approved):

aliculPix4D (approved):

resource/resource_test.go
```diff
@@ -309,6 +309,62 @@ func TestOutE2E(t *testing.T) {
...
+			want:         "https://ci.example.com/teams/devs/pipelines/magritte/jobs/paint/builds/42?vars=%7B%22branch%22%3A%22stable%22%2C%22foo%22%3A%22bar%22%7D",
```
aliculPix4D, line 331: For my understanding: what is the meaning of the second variable? How many pipelines we have in this case two (`stable` and `bar`) or three (`master`, `stable` and `bar`)?

aoldershaw: Suppose you want to set an instanced pipeline for each PR to a repo, and you want to test each PR against multiple Go versions. You may have pipelines with multiple instance vars in that case: `{"number": 123, "go": "1.17"}`. Or maybe you want to group all PRs across a set of repos, in which case you might have `{"repo": "cogito", "number": 123}`.

In general, you may want an N-dimensional "matrix" of instanced pipelines

PR #48, opened 2021-09-14
CI GitHub actions part 2: enable linting, deprecate log_url

marco-m-pix4d:
- doc: update contributing and development documentation
- break: remove debug logging to Google Chat, deprecate `log_url`
  - Since Concourse 7.x _finally_ prints logging output also for the `check` step, we removed support for logging the output of the check step to Google Chat: now all output is printed to the web UI and to `fly check-rersource`.
  - The `log_url` attribute for the source configuration is **DEPRECATED**, will be removed in a future release and is a no-op.
  - Funny fact: I realised I could remove this code attempting to make the linter happy, because the log code was not checking the error status! The linter is your friend :-)
- ci: enable linting and fix current linting errors

PCI-2030


.golangci.yml
```diff
@@ -0,0 +1,7 @@
+# See https://golangci-lint.run/usage/configuration/#config-file
+
```
marco-m-pix4d, line 2: Introducing this cfg file allows to simplify Taskfile and especially to ensure that everybody using vscode will pick up the correct linting configuration, as long as you have

```
  "go.lintTool": "golangci-lint",
  "go.lintFlags": [
    "--fast"
  ],
```

.golangci.yml
```diff
@@ -0,0 +1,7 @@
...
+linters:
+  # NOTE: `enable` _adds_ to the default linters!
+  # To see the currently enabled linters, run `golangci-lint linters`
+  enable:
+    - gocritic
```
aliculPix4D, line 7: Did we loose `gofmt` check?

resource/resource.go
```diff
@@ -118,16 +117,9 @@ func (r *Resource) Check(
...
-	// `fly check-resource` do NOT show anything printed to stderr unless the `check` executable
-	// itself exited with a non-zero status code :-(
-
-	// Optional. If it doesn't exist or is not a string, we will not log.
-	logURL, _ := source["log_url"].(string)
-	hlog.Infof(logURL, BuildInfo())
-	hlog.Debugf(logURL, "check: started")
-	defer hlog.Debugf(logURL, "check: finished")
+	log.Infof(BuildInfo())
+	log.Infof("check: started beppe maniglia")
```
aliculPix4D, line 121: `beppe maniglia`??

aliculPix4D (approved):

.golangci.yml
marco-m-pix4d: No, it is among the default linters (try with `golangci-lint linters`), but thanks for asking!

aliculPix4D: I am confused by `golangci-lint run --enable gofmt,gocritic ./...`. Why did we have it here before? By mistake?

marco-m-pix4d: Because we didn't have before file `.golangi.yml` and because, honestly, I don't know. I _think_ it was because when I added golangci-lint, gofmt was not a default, but maybe I was wrong all the time...

resource/resource.go
odormond: Ah. Confused him with Beppe Grillo. :-D

odormond (approved):

resource/resource.go
marco-m-pix4d: fixed

PR #49, opened 2021-09-16
ci: add support to make a release, triggered by a git tag

marco-m-pix4d:
It becomes complicated quickly :-/

At the end, I decided to use only one GitHub CI (let's call it GH CI for short) workflow and not to run CI at all for external PRs. I think that what this PR proposes is good enough to start with. If we become more experts with GH CI or if we find something lacking I am OK to change this approach.

I tried having two GH CI workflows, one triggered by commits and one triggered by tags, but I had to duplicate all the code and I really didn't like where I was going.

I considered also having two GH CI workflows, one triggered by `push` and one triggered by `release` events, where `release` is a GitHub release, but I saw two problems:
- same code duplication as mentioned above
- for Cogito, we don't really need the whole machinery of a GH "release": when we do a release, the only artifact is a Docker image in Docker hub.

So I kept the release procedure the same as with Travis:

- pushing a git tag of the form `v1.2.3` will be considered as a release.
- the CI will push the same Docker image with 2 tags: `1.2.3` (so the git tag without the `v` prefix) and `latest`.

It is not possible to enforce that the git tag is made on the master branch, since on CI the repository will be checked out in detached HEAD state, thus no branch information.

It would be possible to use something around `git branch -a --contains tags/<TAG>` to extract a list of branches that contain the tag, but I am not sure how fragile it would be (corner cases), so I did nothing.

This PR also

- removes all Travis leftovers
- adds some protection in the Taskfile, for example it will refuse to push a Docker image if the current commit has a release tag, unless Task is running in CI.

## Some interesting CI runs 

- This is the CI run for this PR. No tag, so only the Docker image `cogito:ci-github-actions-3` is pushed: https://github.com/Pix4D/cogito/actions/runs/1242242057
- This is the CI run for a test repository I created, where there is also a release tag. You can see the simulated docker push in the step `task docker:maybe-push-release: https://github.com/Pix4D/understand-gh-actions/runs/3621442619

PCI-2030


ci/git-branch-or-tag.sh
```diff
@@ -0,0 +1,31 @@
+#! /bin/bash
```
marco-m-pix4d, line 1: I am not very proud of the duplication in the two shell scripts, open to suggestions to reduce duplication, if done in a way that doesn't add complexity.

aliculPix4D (approved):

Taskfile.yml
```diff
@@ -100,22 +102,23 @@ tasks:
...
+  docker:maybe-push-release:
+    desc: If a release tag has been detected, Docker push with the 'latest' tag.
+    cmds:
+      - docker tag {{.DOCKER_FULL_NAME}} {{.DOCKER_BASE_NAME}}:latest
+      - docker push {{.DOCKER_BASE_NAME}}:latest
+    preconditions:
+      - sh: test -n "$CI"
+        msg: This target must run only on CI, not locally.
+    status:
+      - test -z "{{.IS_RELEASE}}"
```
odormond, line 121: For my own education, does the `status` means it will consider the task done when it's not a release (IS_RELEASE is an empty string) and so do nothing, while using a precondition would actually make the task fail?

marco-m-pix4d: Correct. Since the logic to detect if we are doing a release or not comes from Task and not from the workflow, I need this target to be idempotent and safe to trigger at any time.

ci/git-branch-or-tag.sh
odormond: You could replace the `echo` with two functions that would echo their argument or nothing for one and its argument or "TRUE" for the other based on an argument to the script. The script could be called `git-branch-tag-or-release.sh`.

odormond (approved):

ci/git-branch-or-tag.sh
marco-m-pix4d: I was thinking along these lines, but I was cooked last week. I will give it a try.

PR #53, opened 2022-01-03
doc: CONTRIBUTING: explain how to update a Github Action CI secret

marco-m-pix4d:


CONTRIBUTING.md
```diff
@@ -278,6 +279,26 @@ $ task docker-build docker-push &&
     fly -t cogito check-resource-type -r cogito-test/branch:another-branch/cogito
 ```
 
+# Setting up GitHub Action CI
+
+Update the expired secrets as follows.
```
odormond, line 284: How frequently should we expect this to happen?

odormond (approved):

aliculPix4D (approved):

Esysc (approved):

CONTRIBUTING.md
marco-m-pix4d: The answer is at line 289 below: 90 days.

The reason I choose 90 days is because it is the biggest amount in the drop-down list that is pre-filled (so it is faster to select). In the drop-down list there is also a custom value, or never expire.

We can choose something else for sure, for example 6 months, but I am not comfortable with a never expire token, also because it is a PAT instead of being a secret valid only for a specific repo...

odormond: 90 days is fine for me. I was concerned about our ability to quickly distinguish between an expected failure due to the expiry or some weirder reason like changed repo permissions or what not.

marco-m-pix4d: Ah, I see. If you go to the PAT page of the person who created the token, if the token is expired it is clearly marked as such.
On the other hand, in the secrets page of this repo, there will be no indication that the secret has expired, since it is just an opaque env var....

PR #54, opened 2022-01-07
cosmetic: reformat code base to max line lenght 99

marco-m-pix4d:
I also added Task target `lint:linelength`, but it is not hooked to CI because
the tool used sometimes makes wrong decisions. In this PR, I ran it and then decided
how to approach each of the suggested changes.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -22,8 +22,9 @@ vars:
     sh: "echo $(which gotestsum 2> /dev/null)"
   TESTRUNNER: "{{if .GOTESTSUM}}{{base .GOTESTSUM}} --{{else}}go test{{end}}"
   #
-  GOLANGCI_VERSION: v1.42.1
+  GOLANGCI_VERSION: v1.43.0
   GOTESTSUM_VERSION: v1.7.0
+  GOLINES_VERSION: v0.7.0
 
 tasks:
 
@@ -32,11 +33,17 @@ tasks:
     cmds:
       - go install github.com/golangci/golangci-lint/cmd/golangci-lint@{{.GOLANGCI_VERSION}}
       - go install gotest.tools/gotestsum@{{.GOTESTSUM_VERSION}}
+      - go install github.com/segmentio/golines@{{.GOLINES_VERSION}}
 
   lint:
     desc: Lint the code.
     cmds:
       - golangci-lint run ./...
+
+  lint:linelength:
+    desc: Enforce max line length. Use the tool output only as a hint, sometimes it makes wrong decisions.
+    cmds:
+      - golines --max-len=99 --write-output .
 
   test:unit:
     desc: Run the unit tests.
--- a/resource/resource_test.go
+++ b/resource/resource_test.go
@@ -95,7 +95,9 @@ func TestIn(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			r := Resource{}
 
-			version, metadata, err := r.In("/tmp", defSource, oc.Params{}, tc.inVersion, defEnv, silentLog)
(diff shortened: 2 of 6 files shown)


Taskfile.yml
```diff
@@ -32,12 +33,18 @@ tasks:
...
 
   lint:
     desc: Lint the code.
     cmds:
       - golangci-lint run ./...
 
+  lint:linelength:
+    desc: Enforce max line length. Use the tool output only as a hint, sometimes it makes wrong decisions.
+    cmds:
+      - golines --max-len=90 --write-output .
```
odormond, line 46: For python code, we use 99. I think it would be simpler to have the same limit independently of the language.

help/testhelper.go
```diff
@@ -47,7 +47,7 @@ func IdentityRenamer(name string) string {
 // └── foo
 //     └── .git        <= dot renamed
 //         └── config
-func CopyDir(dst string, src string, dirRenamer Renamer, templatedata TemplateData) error {
+func CopyDir(dst string, src string, dirRenamer Renamer, tmpldata TemplateData) error {
```
odormond, line 50: With the limit set at 99, you could keep the sane name, couldn't you?

resource/resource_test.go
```diff
@@ -95,7 +95,8 @@ func TestIn(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			r := Resource{}
 
-			version, metadata, err := r.In("/tmp", defSource, oc.Params{}, tc.inVersion, defEnv, silentLog)
+			version, metadata, err := r.In("/tmp", defSource, oc.Params{},
+				tc.inVersion, defEnv, silentLog)
```
odormond, line 99: It looks like I'm now really appreciating the way Black formats python code. The following would be much better IMO:
```suggestion
			version, metadata, err := r.In(
				"/tmp", defSource, oc.Params{}, tc.inVersion, defEnv, silentLog
			)
```

Taskfile.yml
marco-m-pix4d: I have to agree I made the error of accepting the 99...
OK, I will redo this thing then.

resource/resource_test.go
marco-m-pix4d: The problem is that this is not enforceable by any tool, so it would drift.
The root cause is that gofmt, on purpose, leaves the decision about line length to the author, and 3rd-party tools are not good enough :-(
Also, I don't particularly like this specific Black decision, the power of gofmt is that it comes with the language and there should be no discussions about its output, but the line length kicks in :-(

odormond (approved):
I'm a bit confused as I thought Go had already strict formatting rules in place at the compiler level.

As I've never been to much in favour of strict formatting, I will approve this PR but I do think that 99 would be a better limit then 90.

resource/resource_test.go
odormond: I understand. Doing the split between the function name and its argument is one of the decision that Black took that I really appreciate. It's one of the thing that really make the code more readable. I'm less happy about it's willingness to split the arguments one per line if they don't all fit on the same line.

Taskfile.yml
odormond: Thanks!

marco-m-pix4d: Done (force push)

resource/resource_test.go
marco-m-pix4d: At the end I followed this Back thing.

Esysc (approved):

odormond (approved):

aliculPix4D (approved):

PR #55, opened 2022-01-10
Update changelog, release v0.6.1

Esysc:


CHANGELOG.md
```diff
@@ -18,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - tests: new Task targets `test:unit` (replaces target `test`) and `test:integration` (replaces target `test-integration`).
 - build: cleaner Taskfile.
 
+### Fixed
+
+- return error text code in case of http code 500.
```
odormond, line 25: The other entries start with an Uppercase letter.
```suggestion
- Return error text code in case of http code 500. 
```

Esysc: Thanks, corrected.

odormond (approved):

CHANGELOG.md
```diff
@@ -18,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - tests: new Task targets `test:unit` (replaces target `test`) and `test:integration` (replaces target `test-integration`).
 - build: cleaner Taskfile.
 
+### Fixed
+
+- Return error text code in case of http code 500.
```
marco-m-pix4d, line 25: A "code" is a number, so we don't really return it in the text :-)
Maybe something like "Return standard error description in addition to the error code"

marco-m-pix4d (approved):

CHANGELOG.md
Esysc: Thanks for that, I'm doing the change.

Esysc:
Force push to squash commits.

PR #56, opened 2022-01-10
github/status test: cleanups

marco-m-pix4d:
Best reviewed commit-per-commit:

- github/status: cosmetic changes (refactor: no behavioral changes)
- github/status test: refactor: split tests in success and failure 
  This simplifies the tests and makes it easier to spot errors.
- github/status test: fix bug in test

Part of PCI-2267

### DIFF
--- a/github/status_test.go
+++ b/github/status_test.go
@@ -15,48 +15,50 @@ import (
 	"github.com/Pix4D/cogito/help"
 )
 
-func TestGitHubStatusUseMockAPI(t *testing.T) {
+func TestGitHubStatusMockAPI(t *testing.T) {
 	cfg := help.FakeTestCfg
 	context := "cogito/test"
 	targetURL := "https://cogito.invalid/builds/job/42"
 	desc := time.Now().Format("15:04:05")
 	state := "success"
+
 	var testCases = []struct {
 		name       string
+		body       string
 		wantErr    string
 		wantStatus int
-		write      string
 	}{
 		{
-			"No errors",
-			"",
-			http.StatusCreated,
-			"Anything goes...",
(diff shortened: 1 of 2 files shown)


odormond (approved):

github/status_test.go
```diff
@@ -15,57 +15,93 @@ import (
...
 		{
-			"Server error",
-			http.StatusText(500),
-			http.StatusInternalServerError,
-			"Something bad happened!",
-		},
-		{
-			"Repo not found",
-			fmt.Sprintf(`
+			name: "Repo not found",
```
Esysc, line 75: I'm wondering now if this is the correct name for this test. `not found` may hide insufficient rights. I don't have a good alternative here, may be `Not found or wrong token`?

Esysc (approved):

github/status_test.go
marco-m-pix4d: I see. Thinking about it, it seems that it is the "repo" name that is misleading. 
Taking a step back, the tests are exercising the GitHub Status API and are reacting to the HTTP status code received from the API. So the name could be simply "status: Not Found" or "interpret a 404 Not Found"

The other test, with `{"non existing repo: Not Found", `, is a bit different, because in that case the intent is really that the repo is non existing, as can be seen from the repo name ("non-existing-really"). On the other hand, it is true that probably when I wrote that test, we hadn't encountered yet this bizarre GitHub behaviour. For this other test, I will make a separate PR.

marco-m-pix4d: Renamed as follows:

```
name: "500 Internal Server Error",
name: "404 Not Found (multiple causes)",
```

PR #57, opened 2022-01-10
github/status test: test the whole error message

marco-m-pix4d:
This shows what we are actually printing to the user in case of error.
We can see that what we print is formatted in a non-uniform way depending on the HTTP status code (sometimes fewer newlines) and that in general the error message is still confusing.

Part of PCI-2267

### DIFF
--- a/github/status_test.go
+++ b/github/status_test.go
@@ -5,14 +5,13 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
-	"path"
 	"reflect"
-	"strings"
 	"testing"
 	"time"
 
 	"github.com/Pix4D/cogito/github"
 	"github.com/Pix4D/cogito/help"
+	"github.com/google/go-cmp/cmp"
 )
 
 func TestGitHubStatusSuccessMockAPI(t *testing.T) {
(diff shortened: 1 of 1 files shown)


odormond (approved):

Esysc (approved):

aliculPix4D (approved):

github/status_test.go
```diff
@@ -66,21 +65,23 @@ func TestGitHubStatusFailureMockAPI(t *testing.T) {
...
+			wantErr: `POST %s/repos/fakeOwner/fakeRepo/statuses/0123456789012345678901234567890123456789
 One of the following happened:
-    1. The repo https://github.com/%s doesn't exist
+    1. The repo https://github.com/fakeOwner/fakeRepo doesn't exist
 	2. The user who issued the token doesn't have write access to the repo
 	3. The token doesn't have scope repo:status
-`,
-				path.Join(cfg.Owner, cfg.Repo)),
+ X-Accepted-Oauth-Scopes: [], X-Oauth-Scopes: [], status 404 (Not Found
+fake body
```
marco-m-pix4d, line 83: Here "fake body" is correctly reported

github/status_test.go
```diff
@@ -66,21 +65,23 @@ func TestGitHubStatusFailureMockAPI(t *testing.T) {
 		wantStatus int
 	}{
 		{
-			name:       "500 Internal Server Error",
-			body:       "Something bad happened!",
-			wantErr:    http.StatusText(http.StatusInternalServerError),
+			name: "500 Internal Server Error",
+			body: "fake body",
```
marco-m-pix4d, line 69: Here "fake body" is incorrectly not reported.
The body should always be reported in case of error because it may contain additional troubleshooting help. Subsequent PR will fix this

PR #66, opened 2022-01-20
Support git remote URL with encoded basic auth

marco-m-pix4d:
Best reviewed commit per commit.

- resource: rename and document targetURL.
- resource: rewrite parseGitPseudoURL to use url.Parse from stdlib
  - This gives better error messages, solves #46 in another way and will make it easier to extend to Enterprise Github, where the host is not githib.com
- resource:parseGitPseudoURL: move the failing test to passing test
  - This confirms that now we support a PR resource that puts basic auth in the git URL (see #46).

Part of PCI-2274

This also solves in another way #46.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## UNRELEASED - xxxx-xx-xx
 
+### Added
+
+- Support the case where the git remote URL contains basic auth information (this doesn't happen with the Concourse git resource, but can happen with a PR resource, see #46 for a discussion).
+  Thanks @Mike-Dax for the initial implementation (#46).
+
+### Changed
+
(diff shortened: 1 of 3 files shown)


marco-m-pix4d:
@Mike-Dax I believe that this fixes (in a different way) your PR #46 and I plan to issue a new release tomorrow. I mentioned your PR in the CHANGELOG, thanks.
Could you please test the Docker image corresponding to this brach, with tag `strip-secret-from-url` ?

```
resource_types:
- name: cogito
  type: registry-image
  check_every: 1h
  source:
    repository: pix4d/cogito
    tag: strip-secret-from-url
```

Be warned that we delete Docker images associated to branches (like this one) pretty fast, so please do not rely on it, use it only for testing (for Cogito, we recommend to use tag `latest` for production).

aliculPix4D (approved):
LGTM

Esysc (approved):

Issue #51, opened 2021-10-28
Support for Enterprise Github?


olwenya:
Does this line mean that for Enterprise Github (not github.com) will now work?

https://github.com/Pix4D/cogito/blob/50a9d9656fed99e13f7c83fc3487e3c3500ef687/resource/resource.go#L378

```
fetching pix4d/cogito@sha256:cef4adfe17e9ee5f7c16a4094e435d88857f663f080507dcb6b61e7390b52852
29291e31a76a [==========================================] 2.7MiB/2.7MiB
ff9cc595764e [======================================] 274.7KiB/274.7KiB
60eff1c2ce14 [==============================================] 116b/116b
507e2309724c [==========================================] 5.9MiB/5.9MiB
This is the Cogito GitHub status resource. Tag: 0.6.0, commit: 5d34381a18, build date: 2021-08-26
resource source configuration and git repository are incompatible.
Git remote: "https://notgithub.com/xxx/xxx.git"
Resource config: host: github.com, owner: "xxx", repo: "xxx". wrong git remote
```

marco-m-pix4d:
Looking into this, there are two parts involved:

- The API endpoint, that should be `https://{hostname}/api/v3` according to [Getting started with the REST API](https://docs.github.com/en/enterprise-server/rest/guides/getting-started-with-the-rest-api).
- The Git remote, that should be `https://{hostname}/{owner}/{repo}.git`.

marco-m-pix4d:
@olwenya could you please confirm that the only difference is the `{hostname}`, both to access the API and to refer to the git remote ? See https://github.com/Pix4D/cogito/issues/51#issuecomment-1012300930

wanderanimrod:
> @olwenya could you please confirm that the only difference is the {hostname}, both to access the API and to refer to the git remote ? See https://github.com/Pix4D/cogito/issues/51#issuecomment-1012300930

@marco-m-pix4d , this is correct.

wanderanimrod:
@marco-m-pix4d , FYI
I cloned this repo, replaced all the `github.com` references with our enterprise GitHub domain, built an image and tested it. It worked! I haven't pushed the changes to my fork because the tests are broken, and I need to parameterize the GH host instead of hard coding it in. 

Do you think the host should become another config option on the resource's `source` (defaulting to `github.com`)?

PR #69, opened 2022-02-25
Add support for GitHub Enterprise (GHE)

wanderanimrod:
**⚠️  DRAFT**
_Tests are still failing due to some hardcoded "github.com" values in them. I am still trying to fix it, but I wanted you to take a look at my approach to adding GHE support first_

## Why
Currently, this resource only supports setting status on repos hosted in public GitHub. The domain `github.com` and `api.github.com` were hardcoded into the resource before this PR.

## What
1. Parameterize public GH hostname from the resource and add it to optional source params as `github_host`. Default to `github.com`, so current users of the resource won't have to change anything about their `source` configs when they use the upgraded resource. 
2. Update the README to point to the new `github_host`
3. (minor) Mention in the README that users can set `check_every` to `never` in addition to setting it to long periods. 

## Tests
- I have tested this against GitHub Enterprise version 2.22.26 (with the `github_host` source param set to my company's GitHub hostname). 
- I have also tested it against public GitHub (without using the `github_host` source param) and it works. 
- It should also work against later versions of GHE (`3.x+`) because public GitHub is on version `3.x`

## TODO
1. Fix failing unit tests
2. Re-enable tests in Dockerfile.
3. Add a test to explicitly test that the resource works with custom `github_host`.

### Note
ℹ️  _This is my very first time writing go-lang, so please let me know about any rookie mistakes/anti-patterns in my changes 😄_

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,8 +21,8 @@ RUN go mod download
 #
 COPY . .
 
-RUN go test ./...  && \
-    go install \
+# RUN go test ./...  && \ 
+RUN go install \
         -ldflags "-w -X 'github.com/Pix4D/cogito/resource.buildinfo=$BUILD_INFO'" \
         ./cmd/check \
         ./cmd/in \
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ With reference to the [GitHub status API], the `POST` parameters (`state`, `targ
 
 - `context_prefix`: The prefix for the context (see section [Effects on GitHub](#effects-on-github)). If present, the context will be `context_prefix/job_name`. Default: empty. See also the optional `context` in the [put step](#the-put-step).
 - `log_level`: The log level (one of `debug`, `info`, `warn`, `error`, `silent`). Default: `info`.
-- `github_host`: The host name of the github instance (to support GitHub Enterprise (GHE)). The resource has only tested against GHE version 2.x, but should work for version 3.x too. Default: `github.com`
+- `github_host`: The host name of the github instance (to support GitHub Enterprise (GHE)). The resource has only been tested against GHE version 2.x, but should work for version 3.x too. Default: `github.com`
 - `log_url`. **DEPRECATED, no-op, will be removed** A Google Hangout Chat webhook. Useful to obtain logging for the `check` step for Concourse < 7.
 
 ## Suggestions
--- a/README.md
+++ b/README.md
@@ -123,11 +123,12 @@ With reference to the [GitHub status API], the `POST` parameters (`state`, `targ
 
 - `context_prefix`: The prefix for the context (see section [Effects on GitHub](#effects-on-github)). If present, the context will be `context_prefix/job_name`. Default: empty. See also the optional `context` in the [put step](#the-put-step).
 - `log_level`: The log level (one of `debug`, `info`, `warn`, `error`, `silent`). Default: `info`.
+- `github_host`: The host name of the github instance (to support GitHub Enterprise (GHE)). The resource has only tested against GHE version 2.x, but should work for version 3.x too. Default: `github.com`
 - `log_url`. **DEPRECATED, no-op, will be removed** A Google Hangout Chat webhook. Useful to obtain logging for the `check` step for Concourse < 7.
 
 ## Suggestions
 
-We suggest to set a long interval for `check_interval`, for example 1 hour, as shown in the example above. This helps to reduce the number of check containers in a busy Concourse deployment and, for this resource, has no adverse effects.
+We suggest to set a long interval for `check_interval`, for example 1 hour or `never`, as shown in the example above. This helps to reduce the number of check containers in a busy Concourse deployment and, for this resource, has no adverse effects.
 
 # The check step
 
--- a/resource/resource.go
+++ b/resource/resource.go
@@ -186,7 +186,7 @@ func (r *Resource) Out(
 		context = fmt.Sprintf("%s/%s", prefix, context)
 	}
 
-	apiServer := fmt.Sprintf("https://api.%s", github.HOST)
+	apiServer := fmt.Sprintf("https://api.%s", githubHost)
 
 	status := github.NewStatus(apiServer, token, owner, repo, context)
 
--- a/resource/resource.go
+++ b/resource/resource.go
@@ -48,9 +48,10 @@ var (
 	}
 
 	optionalSourceKeys = map[string]struct{}{
-		"log_level":      {},
-		"log_url":        {},
-		"context_prefix": {},
+		"log_level":       {},
+		"log_url":         {},
+		"context_prefix":  {},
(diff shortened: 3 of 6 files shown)


Dockerfile
```diff
@@ -21,8 +21,8 @@ RUN go mod download
 #
 COPY . .
 
-RUN go test ./...  && \
-    go install \
+# RUN go test ./...  && \
```
wanderanimrod, line 24: This will be re-enabled once I figure out how these tests work and fix them.

README.md
```diff
@@ -123,11 +123,12 @@ With reference to the [GitHub status API], the `POST` parameters (`state`, `targ
...
 - `log_level`: The log level (one of `debug`, `info`, `warn`, `error`, `silent`). Default: `info`.
+- `github_host`: The host name of the github instance (to support GitHub Enterprise (GHE)). The resource has only tested against GHE version 2.x, but should work for version 3.x too. Default: `github.com`
 - `log_url`. **DEPRECATED, no-op, will be removed** A Google Hangout Chat webhook. Useful to obtain logging for the `check` step for Concourse < 7.
 
 ## Suggestions
 
-We suggest to set a long interval for `check_interval`, for example 1 hour, as shown in the example above. This helps to reduce the number of check containers in a busy Concourse deployment and, for this resource, has no adverse effects.
+We suggest to set a long interval for `check_interval`, for example 1 hour or `never`, as shown in the example above. This helps to reduce the number of check containers in a busy Concourse deployment and, for this resource, has no adverse effects.
```
wanderanimrod, line 131: Is there any problem with recommending `check_every: never`?

resource/resource.go
```diff
@@ -180,7 +186,12 @@ func (r *Resource) Out(
 		context = fmt.Sprintf("%s/%s", prefix, context)
 	}
 
-	status := github.NewStatus(github.API, token, owner, repo, context)
+	apiServer := fmt.Sprintf("https://api.%s", github.HOST) // default
+	if hostname, ok := source["github_host"].(string); ok {
+		apiServer = fmt.Sprintf("https://api.%s", hostname)
```
wanderanimrod, line 191: If someone uses the `source.github_host` param and sets it to say `github.company.com`, this code assumes that their API will be hosted at `api.github.company.com`. This is how most GHE instances I have encountered are set up, and that's how public GitHub is also set up.  I didn't want to introduce two new source params, one for the API, and another for the host.

Should a target GHE's API be hosted on a completely different domain, this code will break. I am open to suggestions on how this can be improved.

wanderanimrod:
🆘 
I am not making fast enough progress with the tests (fixing them and adding a new one for the `github_host` param). I'd appreciate help from one of the reviewers. Thanks.

resource/resource_test.go
```diff
@@ -636,7 +636,7 @@ func TestCheckRepoDirSuccess(t *testing.T) {
 		defer teardown(t)
 
 		t.Run(tc.name, func(t *testing.T) {
-			err := checkRepoDir(filepath.Join(inDir, tc.dir), wantOwner, wantRepo)
+			err := checkRepoDir(filepath.Join(inDir, tc.dir), wantOwner, wantRepo, "github.com")
```
wanderanimrod, line 639: Not sure this is the best way to do this. It is strange that all the other params to `checkRepoDir` are variables, but "github.com" is a literal string...

wanderanimrod:
Linking to this related issue https://github.com/Pix4D/cogito/issues/51

marco-m-pix4d:
Hello @wanderanimrod, it is a pleasure to see a PR description like this one, thanks!
I will try to have a look as soon as possible.

resource/resource_test.go
marco-m: Note 1:
We need to update also the 3 helper functions - sshRemote(), httpsRemote() and httpRemote().
Or maybe we get rid of them and inline the sprintfs.

Note 2:
> Not sure this is the best way to do this. It is strange that all the other params to checkRepoDir are variables, but "github.com" is a literal string...

Indeed. We need to add a field `githubHost` to the struct, after line 610 and initialize it in each test.
We also need to add a few tests with a custom githubHost, to test the feature added in this PR.

marco-m (commented):
Besides the inline comments, we need to add subtests also to `TestOutMockSuccess`, see as example the subtest `name: "source: optional: context_prefix"`, this will make you change also `optionalSourceKeys`.

But I think that it will be way better if you manage to run the tests; what is the exact procedure you follow and what are the errors?

olwenya:
Just wanted to make sure this doesn't go stale. What the current state of this PR?

marco-m-pix4d:
> Just wanted to make sure this doesn't go stale. What the current state of this PR?

I was waiting for @wanderanimrod for a confirmation that all comments have been addressed.

wanderanimrod:
@olwenya , @marco-m 
The past few weeks have been very busy. I haven't had a chance to look at this in a while. The last problem I ran into was still with the tests. The tests use this repo's config (in the `.git` dir) to run, so stubbing all that stuff out to test github enterprise remotes is not trivial. 

I will probably get time to look into again in 2 weeks.

marco-m-pix4d:
@wanderanimrod although I forgot the details, I tried hard to make tests usable by external contributors, so I suggest before digging in to chime back here with a high-level description of what you plan to change, maybe there is a simpler way. Also in case you haven't done yet, have a look at CONTRIBUTING.md, it explains how to setup integration tests.

bluek1te:
Would like to take this over if the original author is planning on no longer working on it. Mocking out the git enterprise seems a bit heavy handed to me though. (You also can't pin it down because the git enterprise version could differ for each instance). For the integration tests, perhaps we could assume that the API is equivalent to github.com and then just swap in the URL for the current mock git repository on the call to the http module.
 
Because at that point I don't think you necessarily care that the URL is still intact (You could just use unit tests for this), but that the payload is compatible with the API.

wanderanimrod:
Thanks @bluek1te . I would appreciate that as I have been unable to pick this up again.

marco-m:
@bluek1te you are welcome to work on this feature. Please note that the code changed a lot after this PR has been opened, it might be simpler just to start from scratch. Could you please start by reading CONTRIBUTING.md and discussing how you would approach this feature and the tests ("talk, then code")? I also feel that CONTRIBUTING.md has become big/confusing, so any suggestions for restructuring the documentation, seen with a pair of "fresh eyes", is welcome!

marco-m-pix4d:
hello @wanderanimrod, do you have time to rebase on current master? Many things have changed since the base branch of this PR.

wanderanimrod:
> hello @wanderanimrod, do you have time to rebase on current master? Many things have changed since the base branch of this PR.

@marco-m , I just got around to doing this finally. The new codebase structure has made the change much easier to implement. I have created a new PR #139 . Going to close this one. Let's move the discussion there.

PR #75, opened 2022-05-13
Speed up Docker build and doc clarifications

marco-m-pix4d:

part of PCI-2071.


aliculPix4D (approved):

CONTRIBUTING.md
```diff
@@ -264,14 +266,20 @@ $ fly -t cogito set-pipeline -p cogito-test -c pipelines/cogito.yml \
...
 $ task docker-build docker-push &&
     fly -t cogito check-resource-type -r cogito-test/branch:stable/cogito &&
     fly -t cogito check-resource-type -r cogito-test/branch:another-branch/cogito
 ```
 
+## Solving Concourse error: version is missing from previous step
+
+![screenshot of version is missing error](doc/version-is-missing.png)
+
+If you get the confusing `version is missing from previous step` error in cogito put, while using a pipeline like `cogito.yml` that uses as Docker tag the name of the branch, it simply means that you forgot to push for the first time the cogito Docker image.
```
aliculPix4D, line 281: I would extend: `push` to `push where` to docker registry? which one?

odormond (approved):

CONTRIBUTING.md
marco-m-pix4d: The push is done via Task, which has a default push location, that can be overridden for somebody not from Pix4D, so I would think it is fine as-is, no? The fact that there is target docker:push is mentioned above in the same document...

PR #79, opened 2022-05-20
Add Google chat as optional sink

marco-m-pix4d:
To answer the open questions, it is helpful if you join the gchat space `cogito-test` and then trigger a few times the test pipeline, with name `cogito-test`.

Open questions:

1. Defaults for chat notification events. Currently there is a 1:1 mapping between GH Commit status and chat notification.
    Do we want to skip `pending` by default ? But why ?
2. Currently we make use of the gchat "thread key", using the pipeline name as thread key. This means that all notifications related to a specific pipeline always go to the same thread. This seems nice in theory, but in practice, given how the gchat UI renders threads (it jumps everywhere and if you have multiple new threads, once you focus the space, all all cleared at once), I am not so sure.
3. Are all the sinks equal? Currently it is enough for one sink (gh commit status or ghat message) to fail for the put operation to fail. This has the big advantage that behavior is easy to understand and one can (should) always protect this kind of non-really-impacting-build Concourse pipeline in a try block.

Part of PCI-2071


marco-m-pix4d:
Sigh for the CI failure.

odormond (approved):
> - Defaults for chat notification. Currently there is a 1:1 mapping between GH Commit status and chat notification.

You mean, the message content? I would say that depends on the answer to the "thread key" question. Overall it should contain all the information you already put in there and I don't see much point in adding noise around them.

> - Do we want to skip pending by default ? But why ?

With pending being notified, we simply double the number of notification, aren't we? And what would the users gain? Note the plural form, 1 user waits for his job to start but many more will be notified in the channel. For me this is really just noise.

> - Currently we make use of the gchat "thread key", using the pipeline name as thread key. This means that all notifications related to a specific pipeline always go to the same thread. This seems nice in theory, but in practice, given how the gchat UI renders threads (it jumps everywhere and if you have multiple new threads, one you focus the space, all all cleared at once), I am not so sure.

IMO, if the webhook targets a "discussion" channel then for sure it'll be a mess no matter what the `threadKey` looks like. So let's consider a channel dedicated to notifications. The `threadKey` can then be built in different ways:
- `threadKey := env.Get("BUILD_PIPELINE_NAME")` (current code), we get a single thread per pipeline. The thread will become huge over time but gchat will only show the last few messages in it (useful) and the first one (useless). Finding the status of a particular job might require to click on the fold and then scroll back and finding the right message will be difficult as all the jobs are interleaved (bad). Jumping can be very limited if there is just a few pipelines sending notification to this channel (good). Matching the various status updates can be difficult with jobs running in parallel (bad).
- `threadKey := env.Get("BUILD_PIPELINE_NAME") + " " + env.Get("BUILD_JOB_NAME")`, each job (box) in a pipeline get its own thread and so the status updates are easy to follow (good). Only the last updates are shown (useful) + the first message (useless). As pipelines have many jobs, finding a particular job status might require to scroll back instead of clicking on the fold in the previous example but identifying the jobs is easy as they have clear boundaries (good). Jumping will be higher as jobs thread from various pipelines get interleaved (but have clear boundaries) (bad). The job that runs last will always be at the bottom anyway (good).
- `threadKey := env.Get("BUILD_PIPELINE_NAME") + " " + env.Get("BUILD_JOB_NAME") + " " + env.Get("BUILD_NAME")`, each build of each job of each pipeline get a thread. So build boundaries are clear and status matching is trivial (good). Finding older builds requires a lot of scrolling and attention as all build of all jobs of all pipelines get interleaved (bad).
- no `threadKey`, I don't really know what happens then but I expect either all the messages are in a single thread or each is its own thread. In both cases matching status updates will be very difficult (bad).

Personally, I would prefer `BUILD_PIPELINE_NAME + BUILD_JOB_NAME`.

> - Are all the sinks equal? Currently it is enough for one sink (gh commit status or ghat message) to fail for the put operation to fail. This has the big advantage that behavior is easy to understand and one can (should) always protect this kind of non-really-impacting-build Concourse pipeline in a try block.

Well, I wouldn't say they are equal. GH status can block a merge but not the chat message. Now that we have enabled automatic retry, I would even argue that the resource should never fail and instead error if it cannot update the GH status (I expect this to be safe to retry) but I don't know if this would really be possible. If not, then the gchat status should be ignored and only the GH status update should matter for the operation.

marco-m-pix4d:
@odormond I will then answer more in details to your comments.
I had an idea regarding thread key: at the end, we would like to emulate what Concourse cannot give us: all the events associated to one pipeline "run". Instead, it gives us dissociated "job builds" (a pipeline is just a bag of jobs).

But now, the chat notification is generated by cogito, which always has at its disposal a git commit hash. So we could use the hash as thread key. Granted, it would be possible to get multiple job builds with the same commit, with different reasons depending on the pipeline shape, but maybe it would be still better as thread key than pipeline or pipeline + job 🤔

odormond:
> @odormond I will then answer more in details to your comments. I had an idea regarding thread key: at the end, we would like to emulate what Concourse cannot give us: all the events associated to one pipeline "run". Instead, it gives us dissociated "job builds" (a pipeline is just a bag of jobs).
> 
> But now, the chat notification is generated by cogito, which always has at its disposal a git commit hash. So we could use the hash as thread key. Granted, it would be possible to get multiple job builds with the same commit, with different reasons depending on the pipeline shape, but maybe it would be still better as thread key than pipeline or pipeline + job thinking

I like this idea of using the commit for the threadKey! That would be really nice!

resource/chatadapter_test.go
```diff
@@ -0,0 +1,54 @@
...
+		{
+			pipeline: "coconut",
+			state:    "failure",
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc // Capture range variable, needed for t.Parallel to work correctly.
+		t.Run(tc.pipeline, func(t *testing.T) {
+			t.Parallel()
```
aliculPix4D, line 41: This ensures these tests cases are run in parallel?

resource/chatadapter_test.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"context"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/Pix4D/cogito/resource"
+)
+
+func TestChatAdapterMock(t *testing.T) {
```
aliculPix4D, line 12: Can you explain what this is? and why its empty?

aliculPix4D (approved):
> 1) Do we want to skip pending by default ? But why ?

For me this is just a noise and its quite different from the today behavior and usage of Gchat resource.
To go even further, I would even make `on_success` optional and turned off by default. This should be configurable on put step level. Because users today, have control if the want to put Gchat message or not on each of the step. We are taking away this option from them.

> 2) I like this idea of using the commit for the threadKey! That would be really nice!

Seems like a good idea. Do you have an example in test Gchat channel already? 
Otherwise, `BUILD_PIPELINE_NAME` + `BUILD_JOB_NAME` seams reasonable also.

3) Are all the sinks equal?

I would say that in 95% of the cases, this failures will have nothing to do with Github or Gchat. They will be due to Concourse missing volume, which should be re-triggered automatically.

What is confusing me here is the meaning of the error when cogito fails.
1) if it fails as a part of regular build plan and Concourse hook is activated
- there is a probability that user will be informed in Gchat channel about this failure
vs 
2) if it fails as a part of Concourse hooks
- if Gchat message is not successful at this point, user will not be notified and the only point where the error will be seen is WebUI of the Concourse pipeline.

Note that I have seen pipelines that mix this two approach for github status and Gchat.

resource/chatadapter_test.go
marco-m-pix4d: In this case, it requires all the subtests (there is a t.Run) to run in parallel. By default tests are run sequentially.

resource/chatadapter_test.go
marco-m-pix4d: Ah, I had forgotten about this :-/

resource/chatadapter.go
```diff
@@ -0,0 +1,53 @@
...
+	case "failure":
+		icon = "🔴"
+	case "error":
+		icon = "🟠"
+	default:
+		icon = "❓"
+	}
+
+	// Unfortunately the font is proportional and doesn't support tabs,
+	// so we cannot align in columns.
```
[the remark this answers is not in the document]
marco-m-pix4d: I considered the fixed font, but it seemed even worse looking. As you point out, there is also the additional problem of the clickable links.

At least for the time being I will keep as-is...

README.md
```diff
@@ -142,8 +175,9 @@ No-op.
 # The put step
 
 Sets the GitHub commit status for a given commit, following the [GitHub Commit status API].
+The same commit can have multiple statuses, differentiated by parameter `context`.
```
aliculPix4D, line 178: but you don't see this in the Github? commit is either pass or fail? except of course for build status?

resource/resource.go
```diff
@@ -50,7 +52,12 @@ var (
 		"log_level":      {},
 		"log_url":        {},
 		"context_prefix": {},
+		//
+		"gchat_webhook": {},
 	}
+
+	// States that will trigger a chat notification.
+	statesToNotifyChat = []string{"error", "failure"}
```
aliculPix4D, line 60: But I would make `success` configurable/optional. There are pipelines (quite a few pipeline actually, but not enough to justify doing it everywhere) where it is wanted to configure also the success one.

Example: when we build Docker image in linux image build. I saw in a lot of channels also successful pertracker job registrations...

README.md
marco-m-pix4d: I am not sure I understand the question :-)
This is just a description of the GitHub API. You can see them in the screenshot that we already have in the README.

README.md
```diff
@@ -154,7 +154,9 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
...
-- `gchat_webhook`. Default: empty. URL of a [Google Chat webhook]. A notification about the build status will be sent to the associated chat channel, using the pipeline name as thread ID. By default the build statuses that will trigger a notification are `errror` and `pending`. See section [Effects on Google Chat](#effects-on-google-chat) (since v0.7.0).
+- `gchat_webhook`. Default: empty. URL of a [Google Chat webhook]. A notification about the build status will be sent to the associated chat channel, using the pipeline name as thread ID. . See section [Effects on Google Chat](#effects-on-google-chat) (since v0.7.0).
+- `gchat_webhook`. Default: empty. URL of a [Google Chat webhook].
+    A notification about the build status will be sent to the associated chat channel, using a thread key composed by the pipeline name and commit hash. By default the build statuses that will trigger a notification are `errror` and `pending`. See section [Effects on Google Chat](#effects-on-google-chat) (since v0.7.0).
```
aliculPix4D, line 159: typo: errror

resource/resource.go
marco-m-pix4d: This is why I had by default also success and pending.

I want to start by keeping the changes as little as possible, this a new feature and we don't even know if having a resource doing two things works in production.

In this particular case, I can see the rationale for making this setting customizable, but we are not there yet.

README.md
marco-m-pix4d: Ah thanks. This is the dying keyboard of my laptop :-/

README.md
marco-m-pix4d: See also for example https://github.com/Pix4D/cogito-test-read-write/commits/stable and click on any X or V sign to see.

README.md
```diff
@@ -124,6 +154,9 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
 ## Optional keys
 
 - `context_prefix`: The prefix for the context (see section [Effects on GitHub](#effects-on-github)). If present, the context will be `context_prefix/job_name`. Default: empty. See also the optional `context` in the [put step](#the-put-step).
+- `gchat_webhook`. Default: empty. URL of a [Google Chat webhook]. A notification about the build status will be sent to the associated chat channel, using the pipeline name as thread ID. . See section [Effects on Google Chat](#effects-on-google-chat) (since v0.7.0).
```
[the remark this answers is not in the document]
marco-m-pix4d: fixed, thanks.

marco-m-pix4d:
@Pix4D/integration

Changes addressed in this PR after discussion:

- Chat notifications are sent only on error and failure. Skipped: pending and success.
- Chat thread key moved to pipeline + commit hash

Changes to be addressed in subsequent PR:

- Decide how to allow customization of chat notification states.
- The whole discussion that was open question 3: 
    > \3. Are all the sinks equal? Currently it is enough for one sink (gh commit status or ghat message) to fail for the put operation to fail.

Because I think there are already many changes in this PR and I want to keep the discussion focused. Thanks for the feedback!

PR #81, opened 2022-05-25
More diagnostics

marco-m-pix4d:
More debug logging: see clearly which features are enabled (GitHub Commit status, Google Chat notification) and the reason for skipping the sending of a chat message.

This was triggered by the fact that I could not see chat messages for our main pipeline until I remembered that it was because on feature branches we set the hook to nil.

TODO

- [x] rename the redactme!
- [x] discussion about key constants

### DIFF
--- a/resource/utils_test.go
+++ b/resource/utils_test.go
@@ -8,7 +8,7 @@ import (
 
 func TestRedact(t *testing.T) {
 	secrets := map[string]struct{}{
-		"redactme!": {},
+		"redactme": {},
 	}
 
 	testCases := []struct {
@@ -19,30 +19,30 @@ func TestRedact(t *testing.T) {
 		{
 			name: "redact one entry",
 			dirty: map[string]any{
-				"redactme!": "the secret",
-				"mango":     42,
+				"redactme": "the secret",
+				"mango":    42,
 			},
 			want: map[string]any{
-				"redactme!": "REDACTED",
-				"mango":     42,
+				"redactme": "REDACTED",
+				"mango":    42,
 			},
 		},
 		{
 			name: "cannot redact: value to redact is not a string",
 			dirty: map[string]any{
-				"redactme!": 1,
+				"redactme": 1,
 			},
 			want: map[string]any{
-				"redactme!": 1,
+				"redactme": 1,
 			},
 		},
 		{
 			name: "empty string is not redacted",
 			dirty: map[string]any{
-				"redactme!": "",
+				"redactme": "",
 			},
 			want: map[string]any{
-				"redactme!": "",
+				"redactme": "",
 			},
 		},
 	}


resource/utils_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	testCases := []struct {
+		name  string
+		dirty map[string]any
+		want  map[string]any
+	}{
+		{
+			name: "redact one entry",
+			dirty: map[string]any{
+				"redactme!": "the secret",
```
aliculPix4D, line 22: Why the exclamation mark in `redactme!` ?

aliculPix4D (approved):

resource/utils_test.go
marco-m-pix4d: To make it more evident; is this confusing somehow ?
What is important is that it matches a key in lines 10--12

aliculPix4D: For a second or two, I asked myself if this with exclamation mark is some kind of Golang trick. Then I had a look and I could not find any other possibility except this being a simple string. So to answer is it confusing, not much, but I would prefer:

`secret_string_to_redact` or something equivalent i.e `SecretStringToRedact`

marco-m-pix4d: The exclamation mark cannot be a Go trick (well, should not be a Go trick from the point of view of a decent API) because that string is part of the public API: any key set in `secrets` will have its value redacted. 

The `redact` documentation says:

```go
// redact makes a copy of dirty and, for each k/v pair of it, if k matches one of the
// keys in secrets AND v is a string, it redacts v. It returns the redacted copy.
// WARNING: it is able to redact only string values!
func redact(dirty map[string]any, secrets map[string]struct{}) map[string]any
```

But I can remove the `!`, it is not needed.

The thing with `secret_string_to_redact` is that it risks being even more confusing: it is not the string itself, it is the associated value. It should be something like `key_with_value_to_redact`, this is why I used `redactme`...

I didn't like using `secrets map[string]struct{}`, but Go doesn't have sets, so this is how to emulate a set.
I could also have used the more evident `secrets []string`, but then I would have had an uglier implementation.

If redact was part of the public API, then I should have used `secrets []string`. Since it is private, I went with implementer simplicity. Maybe all this discussion is telling me to use `secrets []string` in any case.

resource/resource.go
```diff
@@ -43,17 +46,22 @@ var (
 	}
 
 	mandatorySourceKeys = map[string]struct{}{
-		"owner":        {},
-		"repo":         {},
-		"access_token": {},
+		"owner":      {},
+		"repo":       {},
+		access_token: {},
```
odormond, line 51: For my own education, why do you have to keep the quotes form the first two? And same question for the first 3 elements of `optionalSourceKeys` below?

odormond (approved):

resource/resource.go
odormond: Ah. I see. It's really a petty they decided that the good old convention of naming constant in UPPERCASE was too clear and unambiguous to be preserved. :disappointed:

Having a mix is IMO even more confusing. Having constant for all would be much better.

resource/utils_test.go
marco-m-pix4d: Done

resource/resource.go
marco-m-pix4d: Done. It has been a pain, but it is now done once and for all :-)

PR #82, opened 2022-06-28
Distinguish abort state when possible

marco-m-pix4d:
* Support state abort. If the state is abort, we map it to error when using the GitHub Commit status API, and we leave it as-is when sending a chat notification.
* Document explicitly in the README the mapping of Concourse build state to the various sinks.
* This allows to map 100% Concourse build state to chat, and keeps the mapping to GitHub Commit status API as it was before (which was already the maximum possible).

Closes PCI-2580

cogito-gchat


marco-m-pix4d:
This comes from an idea I had after a discussion with @aliculPix4D, when I wanted to send a chat notification by default also on abort for the tomato pipelines.

It is now visible that the cause of the notification is an abort, while before it was lumped together with the "error" state kitchen sink.

odormond (approved):

aliculPix4D (approved):

PR #84, opened 2022-07-18
remove ofcourse 0: new skeleton

marco-m-pix4d:
As mentioned, this is the first of a series of PR stacked one on top of the others, that will be merged to master only once.

Add skeleton cogito executable to the resource Docker image
In the following PRs, the 3 executables (check, in, out) will be
replaced by this file, by symlinking them.

For statically linked binaries like Go, this reduces the size by 2/3,
plus the reduction due to the removal of the ofcourse package.

PCI-2587


cmd/cogito/main.go
```diff
@@ -0,0 +1,27 @@
...
+	// - stderr for logging
+	// See: https://concourse-ci.org/implementing-resource-types.html
+	if err := run(os.Stdin, os.Stdout, os.Args); err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+}
+
+func run(in io.Reader, out io.Writer, args []string) error {
+	command := path.Base(args[0])
```
marco-m-pix4d, line 24: In the following PRs, this will dispatch based on the `command`, which is the name of the symlinked executable.

cmd/cogito/main_test.go
```diff
@@ -0,0 +1,13 @@
+package main
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
```
marco-m-pix4d, line 6: I am finally introducing an assert library, after having tried _many_.
This is quite nice and allows to reduce the verbosity, focusing on the intent.

go.mod
```diff
@@ -6,11 +6,12 @@ require (
...
-	github.com/google/go-cmp v0.5.4
+	github.com/google/go-cmp v0.5.8
 	github.com/sasbury/mini v0.0.0-20181226232755-dc74af49394b
+	gotest.tools/v3 v3.3.0
 )
 
 require (
 	github.com/alexflint/go-scalar v1.0.0 // indirect
-	github.com/stretchr/testify v1.7.0 // indirect
+	github.com/stretchr/testify v1.7.2 // indirect
```
marco-m-pix4d, line 16: "indirect" means that it is required by a dependency, in this case ofcourse. So it will go away at the end of this series of PRs.

resource/chatadapter.go
```diff
@@ -27,7 +27,7 @@ func sendToChat(
 	instanceVars := env.Get("BUILD_PIPELINE_INSTANCE_VARS")
 	buildURL := concourseBuildURL(atc, team, pipeline, job, buildN, instanceVars)
 
-	webhook, ok := source["gchat_webhook"].(string)
```
marco-m-pix4d, line 30: leftovers

odormond (approved):

resource/chatadapter.go
aliculPix4D: Is this the only change needed? `gchat_webhook` is declared in source config. Why does it still work if you use `gchatWebhookKey`?

aliculPix4D (approved):

resource/chatadapter.go
marco-m-pix4d: Because `gchatWebhookKey` has the value `"gchat_webhook"` :-)

aliculPix4D: > Because `gchatWebhookKey` has the value `"gchat_webhook"` :-)

Ok. Now its clear. Thank
https://github.com/Pix4D/cogito/blob/fa8446007ec2314c4768182ca0f2b40945438d6a/resource/resource.go#L27

PR #85, opened 2022-07-18
Remove ofcourse 1: check step

marco-m-pix4d:
check step: remove ofcourse and use new executable in resource image

Previously we had:

    /opt/resource/check
    /opt/resource/in
    /opt/resource/out

Now we have:

    /opt/resource/cogito
    /opt/resource/check -> cogito    (symlink)
    /opt/resource/in
    /opt/resource/out

Where /opt/resource/cogito is the executable introduced in PR #84.

PCI-2587

TODO

- [x] update comment regarding env vars in check step
- [x] why go.mod has alecthomas/repr ?
- [x] check step can return an empty list for version
- [x] create ticket to get rid of the dummy version as suggested by Olivier.


Dockerfile
```diff
@@ -22,7 +22,6 @@ COPY . .
 RUN go test ./...  && \
     go install \
         -ldflags "-w -X 'github.com/Pix4D/cogito/resource.buildinfo=$BUILD_INFO'" \
-        ./cmd/check \
```
marco-m-pix4d, line 25: we don't build the check executable any more, instead we symlink it to the cogito executable (see line 41)

README.md
```diff
@@ -167,7 +167,7 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
 
 ## Optional keys
 
-- `context_prefix`: The prefix for the context (see section [Effects on GitHub](#effects-on-github)). If present, the context will be `context_prefix/job_name`. Default: empty. See also the optional `context` in the [put step](#the-put-step).
+- `context_prefix`: The prefix for the GitHub Commit status API "context" (see section [Effects on GitHub](#effects-on-github)). If present, the context will be `context_prefix/job_name`. Default: empty. See also the optional `context` in the [put step](#the-put-step).
```
marco-m-pix4d, line 170: just minor clarifications

Taskfile.yml
```diff
@@ -114,7 +114,7 @@ tasks:
       - echo '{{.INPUT}}' | docker run --rm --interactive {{.DOCKER_FULL_NAME}} /opt/resource/in dummy
       - echo
     vars:
-      INPUT: '{ "source": { "owner": "foo", "repo": "bar", "access_token": "123"} }'
+      INPUT: '{ "source": { "owner": "foo", "repo": "bar", "access_token": "123", "log_level": "debug"} }'
```
marco-m-pix4d, line 117: So that task `docker:smoke` will print full debugging output.

cmd/cogito/main.go
```diff
@@ -7,21 +7,38 @@ import (
...
-	if err := run(os.Stdin, os.Stdout, os.Args); err != nil {
+	if err := run(os.Stdin, os.Stdout, os.Stderr, os.Args); err != nil {
 		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
 }
 
-func run(in io.Reader, out io.Writer, args []string) error {
+func run(in io.Reader, out io.Writer, logOut io.Writer, args []string) error {
+	log := hclog.New(&hclog.LoggerOptions{
```
marco-m-pix4d, line 28: This introduces https://github.com/hashicorp/go-hclog, which I used in other projects and consider good.

cmd/cogito/main.go
```diff
@@ -7,21 +7,38 @@ import (
...
+		Name:        "cogito",
+		Output:      logOut,
```
marco-m-pix4d, line 30: stderr in production, buffer or discard in tests.

cmd/cogito/main_test.go
```diff
@@ -1,13 +1,113 @@
...
-	err := run(nil, nil, []string{"foo"})
+func TestRunSmokeSuccess(t *testing.T) {
+	type testCase struct {
+		name    string
+		args    []string
+		in      string
+		wantOut string
+	}
 
+	test := func(t *testing.T, tc testCase) {
```
marco-m-pix4d, line 20: This is a new way to write table-driven tests that I got from https://github.com/gotestyourself/gotest.tools/wiki/Go-Testing-Patterns#table-driven-tests.

The idea is to first show the shape of the test input (struct testCase line 13) and just after (line 20 here) _what_ we are testing.

Only after the test function we enumerate the test cases (start at line 31).

cogito/protocol.go
```diff
@@ -0,0 +1,184 @@
+// This file implements the Concourse resource protocol described at
+// https://concourse-ci.org/implementing-resource-types.html
```
marco-m-pix4d, line 2: Concourse passes to the stdin of the check, in, out executables a partially different JSON object.

This file parses, validates and sets default values both of the stdinput and of other metadata such as environment variables.

cogito/utils_test.go
```diff
@@ -0,0 +1,47 @@
...
+	assert.Assert(t, len(s) == 1 || len(s) == 2)
+	want := s[0]
+	if len(s) == 2 {
+		err := mergo.Merge(&want, s[1], mergo.WithOverride)
+		assert.NilError(t, err)
+	}
+	return want
+}
+
+// failingWriter is an io.Writer that always returns an error.
```
marco-m-pix4d, line 42: So easy

resource/resource.go
```diff
@@ -121,22 +121,7 @@ func (r *Resource) Check(
 	env oc.Environment,
 	log *oc.Logger,
 ) ([]oc.Version, error) {
```
marco-m-pix4d, line 123: we cannot remove this method now because it is needed to implement the ofcourse.Resource interface.

Dockerfile
```diff
@@ -38,3 +37,5 @@ RUN apk --no-cache add ca-certificates
 RUN mkdir -p /opt/resource
 
 COPY --from=builder /root/go/bin/* /opt/resource/
+
+RUN ln -s /opt/resource/cogito /opt/resource/check
```
odormond, line 41: You can probably spare a few more bytes by hard-linking it instead of using a symlink. :smile:

cogito/check.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	log = log.Named("check")
+	log.SetLevel(hclog.LevelFromString(ci.Source.LogLevel))
+	log.Debug("started",
+		"source", ci.Source, "version", ci.Version, "environment", ci.Env, "args", args)
+	defer log.Debug("finished")
+
+	// Since there is no meaningful real version for this resource, we return always the
+	// same dummy version (we need to return a non-empty list to keep Concourse happy).
```
odormond, line 33: This "making Concourse happy" comment is wrong.

From https://concourse-ci.org/implementing-resource-types.html#resource-check:
> The list may be empty, if there are no versions available at the source.

When the s3 resource doesn't find anything Concourse is still happy:
```sh
$ fly -t developers resource-versions -r survey-SURVEY-3502-rename-sources/artifacts-windows.s3 --json -c 1
[]
```

cogito/protocol.go
```diff
@@ -0,0 +1,184 @@
...
+		src.LogLevel = "info"
+	}
+
+	return nil
+}
+
+// Version is a JSON object with string fields. Part of the Concourse resource protocol.
+type Version struct {
+	Ref string `json:"ref"`
+}
```
odormond, line 137: This is generally wrong.

From https://concourse-ci.org/implementing-resource-types.html#resource-check:
> `version` is a JSON object with string fields, used to uniquely identify an instance of the resource.

So the only constraint is that the values must be strings. Nothing is said about the presence of specific keys. And it's easy to find example of versions without a `ref`:
```sh
$ fly -t developers resource-versions -r third-party-master/pushprox-amd64.s3 --json -c 1
[
  {
    "id": 2966343650,
    "version": {
      "path": "final/pushprox/master/pushprox-proxy-v0.1.0-41-g9778f0e"
    },
    "enabled": true
  }
]
```

Given you're allowed to return an empty list, I don't think it makes sense to bother with supporting version in any form for cogito.

cogito/check.go
```diff
@@ -0,0 +1,42 @@
...
+func Check(log hclog.Logger, in io.Reader, out io.Writer, args []string) error {
+	ci, err := NewCheckInput(in)
+	if err != nil {
+		return fmt.Errorf("check: %s", err)
+	}
+
+	log = log.Named("check")
+	log.SetLevel(hclog.LevelFromString(ci.Source.LogLevel))
+	log.Debug("started",
+		"source", ci.Source, "version", ci.Version, "environment", ci.Env, "args", args)
```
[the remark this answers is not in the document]
aliculPix4D: but now I see ...

> None of these variables are available to check

odormond (approved):

cogito/check.go
marco-m-pix4d: I saw that part of the documentation and wondered the same.
I _think_ that when I wrote that comment, for the first version of Cogito, Concourse did want a non-empty list. Or maybe I just misunderstood something, I do not remember.

I will revisit this, thanks.

cogito/protocol.go
marco-m-pix4d: I think that part of the confusion comes from the comment. What is part of the protocol is a JSON object with string fields, the fields are indeed opaque for Concourse.

Here the key "ref" is what I added in the first version of Cogito. It is true that written like this is seems that "ref" is part of the protocol. I will clarify.

As in the previous comment, it might be that previous Concourse versions did want something, I cannot remember.

I prefer to proceed as follows:

- First keep this refactoring a real refactoring: external behavior should stay the same.
- Then, once the migration is complete, try to remove the dummy version. One of the things that leave me a bit unease is what happens to a Concourse that has in the database for Cogito already the `{"ref": "dummy"}` version ...

Thanks for the attention to the specs, greatly appreciated @odormond

cmd/cogito/main_test.go
```diff
@@ -1,13 +1,113 @@
...
+		in := strings.NewReader(tc.in)
+		var out bytes.Buffer
+
+		err := run(in, &out, io.Discard, tc.args)
+
+		assert.NilError(t, err)
+		have := out.String()
```
aliculPix4D, line 27: Are you missing an empty line between L26 and L27?
(given your AAA comment?)

cogito/protocol.go
```diff
@@ -0,0 +1,184 @@
...
+// CheckInput is the JSON object passed to the stdin of the "check" executable plus
+// build metadata (environment variables).
+//
+// See https://concourse-ci.org/implementing-resource-types.html#resource-check
+//
+type CheckInput struct {
+	Source Source `json:"source"`
+	// Field Version will be omitted from the first request.
+	Version Version `json:"version"`
+	// Currently no env variables for check; we keep it for uniformity and just in case.
```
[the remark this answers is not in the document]
marco-m-pix4d: Fixed

go.mod
```diff
@@ -3,15 +3,21 @@ module github.com/Pix4D/cogito
 go 1.18
 
 require (
+	github.com/alecthomas/repr v0.1.0
```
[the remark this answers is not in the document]
marco-m-pix4d: Fixed; I had forgotten to run `go mod tidy`, that removed it.

cogito/check.go
marco-m-pix4d: Will be investigated in PCI-2617.

cogito/protocol.go
marco-m-pix4d: Will be investigated in PCI-2617

cmd/cogito/main_test.go
Esysc: If I'm not wrong, in this case they are in the same assert group.

marco-m-pix4d: Arrange lines 21-22
blank line 23
Act line 24
blank line 25
Assert lines 26-28

Seems correct to me

cogito/utils_test.go
```diff
@@ -0,0 +1,47 @@
...
+func toJSON(t *testing.T, thing any) []byte {
+	t.Helper()
+	buf, err := json.Marshal(thing)
+	assert.NilError(t, err)
+	return buf
+}
+
+// fromJSON unmarshals the JSON-encoded data into thing.
+func fromJSON(t *testing.T, data []byte, thing any) {
+	t.Helper()
```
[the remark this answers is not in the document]
aliculPix4D: Update: I checked here: https://about.sourcegraph.com/blog/go/advanced-testing-in-go
and I assume answer is in this list:
```
Never return errors. Just pass in *testing.T and fail.
By not returning errors, usage is much prettier, since you don't have a bunch of visual overhead from error handling code in your tests since error checking is gone.
Used to make tests clear on what they’re testing vs what is boilerplate
Call t.Helper() for cleaner failure output (Go 1.9)
```

cogito/protocol.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// Source is the "source:" block in a pipeline "resources:" block for the Cogito resource.
+type Source struct {
+	//
+	// Mandatory
+	//
+	Owner       string `json:"owner"`
+	Repo        string `json:"repo"`
+	AccessToken string `json:"access_token"` // SENSITIVE
```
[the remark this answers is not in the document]
aliculPix4D: Ok. I think this answers my question here: https://github.com/Pix4D/cogito/pull/85/files#r924616544

but I am still missing, who invokes the string method?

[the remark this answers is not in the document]
marco-m-pix4d: > who invokes the string method and at which point?

String() implements the fmt.Stringer interface; see
- https://go.dev/tour/methods/17
- https://pkg.go.dev/fmt#Stringer

You can also look at the tests with Redact in their name in this PR.

cogito/protocol.go
```diff
@@ -0,0 +1,188 @@
...
+	ContextPrefix string `json:"context_prefix"`
+}
+
+// String renders Source, redacting the sensitive fields.
+func (src Source) String() string {
+	var bld strings.Builder
+
+	fmt.Fprintln(&bld, "owner:         ", src.Owner)
+	fmt.Fprintln(&bld, "repo:          ", src.Repo)
+	fmt.Fprintln(&bld, "access_token:  ", redact(src.AccessToken))
```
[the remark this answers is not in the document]
marco-m-pix4d: If I understand the question correctly, to use the secret value you use it as any other field: `fetchGit(src.AccessToken)`.

Function redact() (just below here) takes a value, not a pointer, so it will return a redacted copy of the string.

aliculPix4D (approved):

aliculPix4D:
Forgot to approved yesterday :)

cogito/protocol.go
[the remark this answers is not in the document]
aliculPix4D: > String() implements the fmt.Stringer interface

Yes, this part I understand (at least I think I do).
What is not clear is at which point its invoked? By `hclog.Logger` or somewhere in the code and I am just blind?
I saw the test, but I am still lacking understanding of this point.

PR #86, opened 2022-07-20
Remove ofcourse 2: provide debug info as soon as possible, defer input validation

marco-m-pix4d:
This changes the logic of previous PR https://github.com/Pix4D/cogito/pull/85 and makes the code a bit more complicated, with the goal of providing more debug information if something goes wrong at input validation time.

This chicken-and-egg problem is due to the fact that logging configuration is passed
too late, at the same time as all the other resource Source configuration, so to give
as much debugging information as possible we need to get the log level as soon as
possible, also if the Source has other errors. This cannot be simplified, we are
working within the limits of the Concourse resource protocol.

PCI-2587

### DIFF
--- a/cogito/check.go
+++ b/cogito/check.go
@@ -19,16 +19,35 @@ import (
 // the requested version if it is still valid.
 //
 func Check(log hclog.Logger, in io.Reader, out io.Writer, args []string) error {
-	ci, err := NewCheckInput(in)
-	if err != nil {
+	var ci CheckInput
+	dec := json.NewDecoder(in)
+	dec.DisallowUnknownFields()
+	if err := dec.Decode(&ci); err != nil {
+		return fmt.Errorf("check: parsing JSON from stdin: %s", err)
+	}
+	ci.Env.Fill()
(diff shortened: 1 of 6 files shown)


cogito/check.go
```diff
@@ -19,16 +19,35 @@ import (
...
-		return fmt.Errorf("check: %s", err)
+	var ci CheckInput
+	dec := json.NewDecoder(in)
+	dec.DisallowUnknownFields()
+	if err := dec.Decode(&ci); err != nil {
+		return fmt.Errorf("check: parsing JSON from stdin: %s", err)
 	}
+	ci.Env.Fill()
 
+	if err := ci.Source.ValidateLog(); err != nil {
```
marco-m-pix4d, line 30: The complication: we need to validate the source configuration twice: now (line 30), so that we can print debug info, and then at line 42 below.

odormond (approved):

Esysc (approved):

aliculPix4D (approved):

PR #87, opened 2022-07-21
Remove ofcourse 3: get

marco-m-pix4d:
Best reviewed commit per commit.

Previously we had:

    /opt/resource/cogito
    /opt/resource/check -> cogito
    /opt/resource/in
    /opt/resource/out

Now we have:

    /opt/resource/cogito
    /opt/resource/check -> cogito
    /opt/resource/in    -> cogito
    /opt/resource/out

Where /opt/resource/cogito is the executable introduced in the skeleton PR.

PCI-2587


cmd/cogito/main_test.go
```diff
@@ -11,21 +11,19 @@ import (
...
 		var out bytes.Buffer
+		var logOut bytes.Buffer
 
-		err := run(in, &out, io.Discard, tc.args)
+		err := run(in, &out, &logOut, tc.args)
 
-		assert.NilError(t, err)
-		have := out.String()
-		assert.Equal(t, have, tc.wantOut)
+		assert.NilError(t, err, "\nout: %s\nlogOut: %s", out.String(), logOut.String())
```
marco-m-pix4d, line 26: The majority of the [assert](https://pkg.go.dev/gotest.tools/v3/assert) functions have the last parameters of the signature the same:
```
func Equal(t testing.T, have, want interface{}, msgAndArgs ...interface{})
```
where the optional
```
msgAndArgs ...interface{}
```
are printed in case of failure, to give more context.

cogito/protocol.go
```diff
@@ -21,6 +21,20 @@ type CheckInput struct {
...
+// build metadata (environment variables).
+//
+// See https://concourse-ci.org/implementing-resource-types.html#resource-in
+//
+type GetInput struct {
+	Source  Source  `json:"source"`
+	Version Version `json:"version"`
+	// Cogito does not support get params; a resource supporting them would have the
+	// following line uncommented:
+	// Params  GetParams `json:"params"`
```
marco-m-pix4d, line 34: I left this commented out field on purpose, to have `GetInput` as a blueprint for writing a resource. The same "blueprint-thinking" can be seen below, for `GetParams`.

odormond (approved):

Esysc (approved):

cogito/protocol.go
```diff
@@ -137,6 +151,19 @@ func (ver Version) String() string {
...
+// Output is the JSON object emitted by the get and put step.
+type Output struct {
+	Version Version `json:"version"`
+	// the Cogito resource doesn't use the optional "metadata" field.
+	// Metadata Metadata `json:"metadata"`
+}
+
+// GetParams is the "params:" block in a pipeline get step for the Cogito resource.
+// Cogito doesn't accept any get step parameters, so to give a slightly better error
+// message form the JSON parsing, instead of declaring it an empty struct, we do not
```
aliculPix4D, line 163: `form` or `for`?

aliculPix4D (approved):

cogito/protocol.go
marco-m-pix4d: @aliculPix4D 

> `form` or `for`?

3rd possibility: `from` :-). Addressed in last PR in this series, thanks.

>  `so to give a slightly better error message from the JSON parsing`.
> Why and how this gives a better msg?

Because in Go (and in any statically typed language), we can leave the compiler or the runtime do the type validation for us. In this case it is at runtime (since the input comes at runtime).

Imagine that the pipeline had a get step for cogito (forget for a moment that get does nothing for cogito):

```yaml
...
get: gh-status
params: {foo: bar}
```

since the get request struct is:

```go
type GetRequest struct {
	Source  Source  `json:"source"`
	Version Version `json:"version"`
	// Cogito does not support get params; a resource supporting them would have the
	// following line uncommented:
	// Params  GetParams `json:"params"`
	Env Environment
}
```

so, no `GetParams` field, it would the JSON parser itself to return an error, no need for our code (as we were doing with `ofcourse`) to do any input validation at _this_ level.

PR #88, opened 2022-07-27
Remove ofcourse 4: put part 1

marco-m-pix4d:
This PR adds a not-yet-complete new /opt/resource/out executable implementing the put step, arriving to the point of processing the "put input" directory containing the git repo and extracting the git commit hash.

Mostly it migrates existing code and existing tests and testdata from directory cogito/resource to directory cogito/cogito, adapting the code and the tests to the new approach.

While doing the migration, I also slightly simplified the tests and increased test code coverage. Currently we have 100% test code coverage for all the steps.

I also added new Taskfile targets:
- `test:smoke`, that does a smoke test of the 3 executables.
- `test:buildinfo` that verifies that the executable contains the build information added at link time.

Better reviewed commit-per-commit.

Part of PCI-2587.

EDIT: please ignore, at the bottom of the PR, the section added by GitHub with name **Unchanged files with check annotations**. That code is currently unused and has not been migrated yet, it does not impact tests nor build.


Taskfile.yml
```diff
@@ -63,21 +63,56 @@ tasks:
     desc: Run the unit tests.
     cmds:
       # One day I will understand how to use -coverpkg=./... :-(
-      - go test -count=1 -short -coverprofile=coverage.out ./...
+      # FIXME - go test -count=1 -short -coverprofile=coverage.out ./...
```
marco-m-pix4d, line 66: This will be recovered in a subsequent PR (see commit that introduced this for details)

Taskfile.yml
```diff
@@ -63,21 +63,56 @@ tasks:
...
       - echo
       - echo '{{.SMOKE_INPUT}}' | ./bin/in dummy-dir
       - echo
+      - rm -rf /tmp/cogito-test
+      - mkdir -p /tmp/cogito-test
+      - >
+        ./bin/copydir cogito/testdata/one-repo/a-repo /tmp/cogito-test --dot 
+        --template repo_url=https://github.com/foo/bar head=dummyHead
+        branch_name=dummyBranch commit_sha=dummySha
+      - echo '{{.PUT_INPUT}}' | ./bin/out /tmp/cogito-test
```
marco-m-pix4d, line 92: I hope I will be able to keep this smoke test of the put step, but it will need to interact with GitHub...

cmd/cogito/main_test.go
```diff
@@ -6,59 +6,66 @@ import (
...
+	var logOut bytes.Buffer
+	inputDir := testhelp.MakeGitRepoFromTestdata(t, "../../cogito/testdata/one-repo/a-repo",
+		testhelp.HttpsRemote("the-owner", "the-repo"), "dummySHA", "dummyHead")
 
-	for _, tc := range testCases {
-		t.Run(tc.name, func(t *testing.T) {
-			test(t, tc)
-		})
-	}
+	err := run(in, &out, &logOut, []string{"out", inputDir})
```
marco-m-pix4d, line 66: I also transformed some tests to the new pattern with the `test()` anonymous function that I already introduced in this PR series; it makes the test _intent_ more evident because now it appears before the (long) list of test cases.

cogito/buildinfo.go
```diff
@@ -0,0 +1,11 @@
+package cogito
```
marco-m-pix4d, line 1: Just moved

cogito/protocol.go
```diff
@@ -9,6 +9,10 @@ import (
 	"strings"
 )
 
+// DummyVersion is the version always returned by the Cogito resource.
+// DO NOT REASSIGN!
+var DummyVersion = Version{Ref: "dummy"}
```
marco-m-pix4d, line 14: We would like this to be a const but it cannot :-/

cogito/put.go
```diff
@@ -0,0 +1,277 @@
+package cogito
```
marco-m-pix4d, line 1: the meat of the PR, mostly migrated.

cogito/put.go
```diff
@@ -0,0 +1,277 @@
...
+		return fmt.Errorf("put: params: %s", err)
+	}
+	log.Debug("", "state", buildState)
+
+	gitHash, err := processInputDir(inputDir, pi.Source.Owner, pi.Source.Repo)
+	if err != nil {
+		return fmt.Errorf("put: processing the input dir: %s", err)
+	}
+	log.Debug("", "git-commit", gitHash)
+
```
marco-m-pix4d, line 72: Next PR will add the missing logic here.

cogito/put_test.go
```diff
@@ -0,0 +1,220 @@
...
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) { test(t, tc) })
+	}
+}
+
+func TestPutPipelineValidationFailure(t *testing.T) {
```
marco-m-pix4d, line 66: These tests validate errors introduced by the user writing the pipeline

cogito/put_test.go
```diff
@@ -0,0 +1,220 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) { test(t, tc) })
+	}
+
+}
+
+func TestPutProtocolFailure(t *testing.T) {
```
marco-m-pix4d, line 119: These tests validate "should not happen" situation: resource protocol errors introduced by Concourse.

go.mod
```diff
@@ -3,20 +3,20 @@ module github.com/Pix4D/cogito
 go 1.18
 
 require (
-	github.com/alexflint/go-arg v1.3.0
+	github.com/alexflint/go-arg v1.4.3
```
marco-m-pix4d, line 6: used only by helper executables, not by the resource.

testhelp/testhelper.go
```diff
@@ -187,3 +188,63 @@ func MergeMap(a, b map[string]any) map[string]any {
 	}
 	return c
 }
+
```
marco-m-pix4d, line 191: as the majority of the code, this also existed already. Moved and added more documentation.

testhelp/testhelper.go
```diff
@@ -27,7 +27,8 @@ func IdentityRenamer(name string) string {
 	return name
 }
 
-// Recursive copy src directory below dst directory, with optional transformations.
+// CopyDir recursively copies src directory below dst directory, with optional
```
odormond, line 30: There is not much point in using this to just copy a directory so maybe it would be better named `templatedir` or `fleshoutdir`?

odormond (commented):
Expanding the large diff of `resource/resource.go` I notices that github is showing some errors:
image
(it's not the only ones and there is more on untouched files at the end of the PR: `resource/chatadapter.go` and `resource/commitstatusadapter.go`)

Are these things to worry about?

marco-m-pix4d:
> Are these things to worry about?

This is what is mentioned in the PR description:

> EDIT: please ignore, at the bottom of the PR, the section added by GitHub with name *Unchanged files with check
> annotations*. That code is currently unused and has not been migrated yet, it does not impact tests nor build.

Just ignore them. Final PR will ensure not to have any.

cogito/protocol.go
marco-m-pix4d: I can, but I should not (code will compile fine; linter will complain). Still, point taken :-/

Dockerfile
```diff
@@ -19,11 +19,10 @@ RUN go mod download
 #
 COPY . .
 
-RUN go test ./...  && \
-    go install \
-        -ldflags "-w -X 'github.com/Pix4D/cogito/resource.buildinfo=$BUILD_INFO'" \
-        ./cmd/cogito \
-        ./cmd/out
+RUN go test ./cogito ./cmd/cogito
+RUN go install \
```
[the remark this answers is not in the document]
marco-m-pix4d: There is no need to split; I did it because this is not the final image (see line 37 below).
Do you see negative consequences?

cogito/put.go
```diff
@@ -0,0 +1,277 @@
...
+	"github.com/hashicorp/go-hclog"
+	"github.com/sasbury/mini"
+)
+
+// Put implements the "put" step (the "out" executable).
+//
+// From https://concourse-ci.org/implementing-resource-types.html#resource-out:
+//
+// The out script is passed a path to the directory containing the build's full set of
+// sources as command line argument $1, and is given on stdin the configured params and
```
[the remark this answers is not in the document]
marco-m-pix4d: Ah yes. Actually this is copied from the Concourse docs, but I agree it is wrong. Will update accordingly.

- [ ] fix comment

odormond:
> > Are these things to worry about?
> 
> This is what is mentioned in the PR description:
> 
> > EDIT: please ignore, at the bottom of the PR, the section added by GitHub with name _Unchanged files with check
> > annotations_. That code is currently unused and has not been migrated yet, it does not impact tests nor build.
> 
> Just ignore them. Final PR will ensure not to have any.

OK but what about the ones in `resource/resource.go`? I guess the actual question is: what is left for "put part 2"? I'm apparently missing the reason for not removing entirely the `resource/resource.go`.

testhelp/testhelper.go
```diff
@@ -1,4 +1,4 @@
-package help
+package testhelp
```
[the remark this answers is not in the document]
marco-m-pix4d: In Go, a test file must be called `something_test.go`, with the `_test` suffix.

In addition, a test file can have two _package_ names:
- the same name as the package, for example `cogito`. This puts the test code in the same namespace of the SUT, and allows to test non-exported symbols.
- the same name as the package, with a `_test` suffix, for example `cogito_test`. This puts the test code in a different namespace, so that it mimics real client code and has access only to exported symbols of the SUT.

So package name `testhelp` cannot mean that it is used to test the `--help` functionality, although I understand it can be confusing if one doesn't know the Go testing conventions.

If you still find it confusing, I have no problems in renaming, just let me know.

testhelp/testhelper.go
marco-m-pix4d: To give some background:

I seem to remember that I called it (at the time of Cogito creation, this is just moved) `CopyDir` because there is no function to copy a directory in Go stdlib, so if it is invoked as

CopyDir with [IdentityRenamer](https://github.com/Pix4D/cogito/blob/ccc4876f2fa19e80db0e25489c2bf67536418c6c/testhelp/testhelper.go#L26) and with empty templatedata, it will actually behave as a plain CopyDir.

Having said this, I am still open to renaming, or maybe making this parametric function as private and exposing two public functions, with clearer names and more focused argument list ?

Dockerfile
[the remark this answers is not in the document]
odormond: No big deal. It might very slightly increase the final image size because it should record an additional layer but unless the `go test` does touch the filesystem it will be an empty one. (I used conditional because I don't know if it can optimize away and empty layer or not.)

testhelp/testhelper.go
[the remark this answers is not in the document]
marco-m-pix4d: Mhh, you put a bug in my hear that it doesn't want to go away :-)
I will rename it.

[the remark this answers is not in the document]
odormond: OK. Fine for me and thanks for the explanation. I'm too much used to the python convention of using `test_` as prefix instead of the go suffix way. No need to change if it's the go way. :-)

Taskfile.yml
```diff
@@ -63,21 +63,56 @@ tasks:
...
       - echo '{{.SMOKE_INPUT}}' | ./bin/check
       - echo
       - echo '{{.SMOKE_INPUT}}' | ./bin/in dummy-dir
       - echo
+      - rm -rf /tmp/cogito-test
+      - mkdir -p /tmp/cogito-test
+      - >
+        ./bin/copydir cogito/testdata/one-repo/a-repo /tmp/cogito-test --dot 
+        --template repo_url=https://github.com/foo/bar head=dummyHead
+        branch_name=dummyBranch commit_sha=dummySha
```
odormond, line 91: @marco-m-pix4d, this is actually the line that triggered my comment on renaming `copydir`. I didn't think of it being used as a function in go code where a std lib copy isn't available. Just renaming the binary would be enough, IMO.

marco-m-pix4d: Ah, good point. Suggestions for the name of this executable? Same as in the previous comment:  `templatedir` or `fleshoutdir`?

cogito/protocol.go
aliculPix4D: For my understanding: why it cannot be a constant?

aliculPix4D (approved):

Taskfile.yml
odormond: Yes. I prefer `templatedir`.

cogito/protocol.go
marco-m-pix4d: Because the compiler, implementing the Go language specs, is not able to.
Making things constant (or immutable) is not as easy as it seems from a purely conceptual point of view.
See for example (although confusing) https://github.com/golang/go/issues/27975.
I wish that Go 1 had the concept of immutable, but now it is too late.

marco-m-pix4d:
@odormond 

> OK but what about the ones in `resource/resource.go`? I guess the actual question is: what is left for "put part 2"? I'm apparently missing the reason for not removing entirely the `resource/resource.go`.

The idea I gas was to remove from resource/resource.go incrementally what I was migrating in this series of PR.
This worked fine until this PR, were we hit this new beta feature from GitHub that I could not see how to switch off.

I think that I will do the following:
- completely remove resource/resource.go now
- keep a copy around, outside of git, so that I can still use it to compare what is left :-/

Before doing this last resort, I will try to modify resource/resource.go to still keep around what I need and not trigger this github "feature".

To directly answer the question
> what is left for "put part 2"? 

The integration of the two sinks: github commit status and ghcat.
The features are in separate files, but the remainings of resource.go integrate them in the put step.
And the tests, in resource_test.go

Esysc (approved):

cogito/put.go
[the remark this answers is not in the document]
marco-m-pix4d: @odormond updated in the last PR of this series, thanks.

testhelp/testhelper.go
marco-m-pix4d: Discussion proceeds in comment thread below https://github.com/Pix4D/cogito/pull/88#discussion_r931944930

Taskfile.yml
marco-m-pix4d: @odormond 
addressed in last PR of the series.

PR #89, opened 2022-08-02
Remove ofcourse 5: put part 2

marco-m-pix4d:
Refactor Put into the Putter interface.
This makes testing with doubles easier and actually makes Put() a generic wrapper that could be used as-is for any Concourse resource: one has only to implement the Putter interface.

Best reviewed commit-per-commit to reduce noise.

As is the case for all the PRs in this series, for the time being I managed to keep 100% test line coverage.

Next PR will add the sink GitHub Commit Status
Next PR will add the sink Google Chat

Part of PCI-2587


cmd/cogito/main.go
```diff
@@ -38,7 +39,8 @@ func run(in io.Reader, out io.Writer, logOut io.Writer, args []string) error {
 	case "in":
 		return cogito.Get(log, in, out, args[1:])
 	case "out":
-		return cogito.Put(log, in, out, args[1:])
+		putter := cogito.NewPutter(github.API, log)
+		return cogito.Put(log, in, out, args[1:], putter)
```
marco-m-pix4d, line 43: classic dependency injection

cmd/cogito/main_test.go
```diff
@@ -47,7 +47,8 @@ func TestRunGetSmokeSuccess(t *testing.T) {
 	assert.NilError(t, err, "\nout: %s\nlogOut: %s", out.String(), logOut.String())
 }
 
-func TestRunPutSmokeSuccess(t *testing.T) {
+func TestRunPutSmokeIntegrationSuccess(t *testing.T) {
+	t.Skip("FIXME REQUIRES SECRETS") // FIXME
```
marco-m-pix4d, line 51: next PR will enable it

cogito/put.go
```diff
@@ -26,91 +45,167 @@ import (
 // Additionally, the script may emit metadata as a list of key-value pairs. This data is
 // intended for public consumption and will make it upstream, intended to be shown on the
 // build's page.
-func Put(log hclog.Logger, in io.Reader, out io.Writer, args []string) error {
-	var pi PutInput
+func Put(log hclog.Logger, in io.Reader, out io.Writer, args []string, putter Putter) error {
```
marco-m-pix4d, line 48: What happens in Put is generic for any Concourse resource.

cogito/put.go
```diff
@@ -26,91 +45,167 @@ import (
...
-		return "", err
-	}
+func (cs GitHubCommitStatusSink) Send() error {
+	cs.Pu.log.Info("GitHubCommitStatusSink: Send")
 
-	gitHash, err := getGitCommit(repoDir)
-	if err != nil {
-		return "", err
-	}
+	// return GhCommitStatusSink(cs.Pu.ghAPI, cs.Pu.log, cs.Pu.Pi, cs.Pu.gitRef)
```
marco-m-pix4d, line 193: Will be enabled in next PR

cogito/put_private_test.go
```diff
@@ -12,46 +12,6 @@ import (
 	"gotest.tools/v3/assert"
 )
 
-func TestProcessInputDirFailure(t *testing.T) {
```
marco-m-pix4d, line 15: This is now tested as part of the public API in put_test.go

cogito/put_test.go
```diff
@@ -14,47 +15,104 @@ import (
...
+		Source: baseSource,
+		Params: baseParams,
+	}
+)
+
+type MockPutter struct {
+	loadConfigurationErr error
+	processInputDirErr   error
+	outputErr            error
+	sinkers              []cogito.Sinker
```
marco-m-pix4d, line 37: This allows to control the mock at init time and specify if and when to return an error.

marco-m-pix4d:
As for the other PR, please disregard the zealous GitHub Action complaining about "Unchanged files with check annotations"

cogito/put.go
```diff
@@ -13,6 +13,25 @@ import (
...
+	Sinks() []Sinker
+	// Output emits the version and metadata required by the Concourse protocol.
+	Output(out io.Writer) error
+}
+
+// Sinker represents a sink: an endpoint to send a message.
+type Sinker interface {
+	// Send posts the information extracted by the Putter to a specific sink.
+	Send() error
+}
```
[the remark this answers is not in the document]
marco-m-pix4d: Actually it is the opposite :-)

To be fully modular, it has to be as in this PR, with a Send() with no parameters and no LoadConfiguration() or similar. 
The next PR will show how it is used, but I will add later to this comment a minimal example.

cogito/put.go
```diff
@@ -26,91 +45,167 @@ import (
 // Additionally, the script may emit metadata as a list of key-value pairs. This data is
 // intended for public consumption and will make it upstream, intended to be shown on the
 // build's page.
-func Put(log hclog.Logger, in io.Reader, out io.Writer, args []string) error {
-	var pi PutInput
+func Put(log hclog.Logger, in io.Reader, out io.Writer, args []string, putter Putter) error {
+	if err := putter.LoadConfiguration(in, args); err != nil {
+		return fmt.Errorf("put: %s", err)
```
Esysc, line 50: Does it make sense here and below adding the caller in the prefix? Example `put LoadConfiguration:...`

cogito/put.go
```diff
@@ -26,91 +45,167 @@ import (
...
+		"environment", pu.Pi.Env,
 		"args", args)
 
-	if err := pi.Source.Validate(); err != nil {
+	if err := pu.Pi.Source.Validate(); err != nil {
 		return fmt.Errorf("put: %s", err)
 	}
 
 	// args[0] contains the path to a directory containing all the "put inputs".
 	if len(args) == 0 {
```
Esysc, line 119: For my understanding, should this check moved on the top?

cogito/put.go
```diff
@@ -26,91 +45,167 @@ import (
...
+		ghAPI: ghAPI,
+		log:   log,
+	}
+}
+
+func (pu *ProdPutter) LoadConfiguration(in io.Reader, args []string) error {
 	dec := json.NewDecoder(in)
 	dec.DisallowUnknownFields()
-	if err := dec.Decode(&pi); err != nil {
+	if err := dec.Decode(&pu.Pi); err != nil {
```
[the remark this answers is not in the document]
marco-m-pix4d: At a certain point you will have to accept what is idiomatic Go and not try to write Python in Go or to think Python while reading Go code :-)

The same as I accept the in my opinion error-prone and added cognitive load of Python functions when one never knows wether they raise an exception or return an error, and the only way to know is to read the whole called function in a recursive never ending way. This is why Go does error handling the way it does it: impossible to forget to handle an exception, since they do not exist. The only thing you need to look at in order to correctly handle a function call is to read the function type :-)

Going into the specifics of the code here:

`Pi` cannot be called a `Request` because it is more than a Request, it also contains the environment variables. This is why I called it `PutInput`. In this case, in my opinion the error is on the Concourse protocol design, since they pass "input" in 3 channels:
1. stdin (the JSON object)
2. command-line arguments (the directory which contains the resource inputs volumes)
3. environment variables (build number, pipeline name etc)
while I think it would have made more sense to encode everything in a JSON object...

We could not even tell ourselves that we consider that these 3 "inputs" are still part of one request, because this request is actually different in the 3 steps, so we would have to call it `PutRequest`, `GetRequest` and `CheckRequest`. Which is what we have now with `PutInput`, `GetInput` and `CheckInput`. I am not satisfied either with the `Input` suffix, since it can be confused with the "input directory".

If you think that here replacing the `Input` suffix with `Request` would make it less-confusing, I am OK with renaming. Once we decide, we could have one the following

- pu.PutInput
- pu.PutRequest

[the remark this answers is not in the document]
odormond: Yes, using `PutRequest`, `GetRequest` and `CheckRequest` would be much better!

cogito/put.go
marco-m-pix4d: Ah, I see the reasoning. It is indeed always confusing to decide the "point of view" when reporting errors. This is what I attempt to do in cogito:

The "LoadConfiguration" prefix (or anyway something clear enough to locate the error) should be added by the LoadConfiguration() function itself when returning an error. Example:

```
func LoadConfiguration() {
    ...
    return fmt.Errorf("parsing stdin: %s", ...)
}
```
where in this case the full error would become:
```
put: parsing stdin: ...
```
and be clear enough.

cogito/put.go
marco-m-pix4d: I think you mean "should this check be moved at the beginning of the function?".
I considered this, and I am not even 100% sure that what I did is the best approach. The reason it is this way has been mentioned in a previous PR I think: it is to be able to know the log level :-(

The log level is passed too late, as part of the "source" configuration. This is why "LoadConfiguration" is a mess: it needs to keep going until it gets the log level, so that it can give as much diagnostic information as possible.

If I were to put this check at the beginning of the function, and the check failed, and the log level where set to debug, we could not print any log, only the error message....

One might ask: why we didn't have this problem before, with "ofcourse" ?
Because ofcourse somehow "cheated": the log level was part of ofcourse, and it was ofcourse that was doing the JSON parsing, and calling Put() with a properly initialized logger.

So one might ask: so what is the advantage of removing ofcourse?
The advantage is type safety and LESS code. Ofcourse was obliged to parse the JSON input as a typeless map string->string (as a Python implementation would do) and leave the input validation (both types and field names) to the implementation (you could compare the current master of cogito). On the other hand, this series of PR is doing it the Go idiomatic way: type safety AND field names are ensured automatically by the fact that we use Go structs with a 1-to-1 mapping to the resource YAML configuration (if you search in this series of PRs for JSON parsing tests, you will see that sometimes I pass raw text with some JSON objects, and the input is refused automatically by the JSON parser, without even reaching the cogito code).

Esysc: Thanks for taking the time it's much clearer now.

cogito/put.go
Esysc: Ah sure, I missed the "point of view".

Esysc (approved):

cogito/put.go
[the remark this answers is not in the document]
marco-m-pix4d: Types: switched to `PutRequest`, `GetRequest` and `CheckRequest`.
Variables: from `ci`, `Pi`, `gi` to `request` (context is clear enough to understand which request it is).

It is now more readable indeed. Thanks!

odormond (approved):

cogito/put.go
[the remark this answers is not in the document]
marco-m-pix4d: Explained in next PR: https://github.com/Pix4D/cogito/pull/90#discussion_r940276017

PR #90, opened 2022-08-08
Remove ofcourse 6: put part 3: add sink GH Commit Status

marco-m-pix4d:
Finally add back the GitHub Commit status API.

Best reviewed commit-per-commit, each commit message explains the changes.

This PR also adds a development dependency, the `tparse` utility, to print a better summary of a test run. Run `task install:deps` to install.


Taskfile.yml
```diff
@@ -25,27 +25,22 @@ vars:
...
       - go install github.com/golangci/golangci-lint/cmd/golangci-lint@{{.GOLANGCI_VERSION}}
-      - go install github.com/segmentio/golines@{{.GOLINES_VERSION}}
+      - go install github.com/mfridman/tparse@{{.TPARSE_VERSION}}
 
   lint:
     desc: Lint the code.
     cmds:
       - golangci-lint run ./...
 
-  lint:linelength:
```
marco-m-pix4d, line 44: Turned out to be useless and buggy.

cmd/cogito/main.go
```diff
@@ -32,14 +32,21 @@ func run(in io.Reader, out io.Writer, logOut io.Writer, args []string) error {
 	})
 	log.Info(cogito.BuildInfo())
 
+	ghAPI := os.Getenv("COGITO_GITHUB_API")
```
marco-m-pix4d, line 35: This is undocumented on purpose and makes sense only when running tests, to pass the mock http server.
It is done here in main.go because it allows to deeply test almost all the code, without having to spawn a subprocess.

See how it is used in `TestRunPutSuccess` in cmd/cogito/main_test.go

cmd/cogito/main_test.go
```diff
@@ -47,8 +51,12 @@ func TestRunGetSmokeSuccess(t *testing.T) {
...
 }
 
-func TestRunPutSmokeIntegrationSuccess(t *testing.T) {
-	t.Skip("FIXME REQUIRES SECRETS") // FIXME
+func TestRunPutSuccess(t *testing.T) {
+	wantState := cogito.StatePending
+	wantGitRef := "dummyHead"
+	var ghReq github.AddRequest
+	var URL *url.URL
+	ts := testhelp.GhCommitStatusTestServer(&ghReq, &URL)
```
marco-m-pix4d, line 59: spawn the test server

cmd/cogito/main_test.go
```diff
@@ -62,14 +70,45 @@ func TestRunPutSmokeIntegrationSuccess(t *testing.T) {
 	var out bytes.Buffer
 	var logOut bytes.Buffer
 	inputDir := testhelp.MakeGitRepoFromTestdata(t, "../../cogito/testdata/one-repo/a-repo",
-		testhelp.HttpsRemote("the-owner", "the-repo"), "dummySHA", "dummyHead")
+		testhelp.HttpsRemote("the-owner", "the-repo"), "dummySHA", wantGitRef)
+	t.Setenv("COGITO_GITHUB_API", ts.URL)
```
marco-m-pix4d, line 74: pass the test server URL to the SUT

cmd/cogito/main_test.go
```diff
@@ -62,14 +70,45 @@ func TestRunPutSmokeIntegrationSuccess(t *testing.T) {
...
 
 	err := run(in, &out, &logOut, []string{"out", inputDir})
 
 	assert.NilError(t, err, "\nout: %s\nlogOut: %s", out.String(), logOut.String())
+	ts.Close() // Avoid races before following asserts.
```
marco-m-pix4d, line 79: As the comment says, this is a "belt and suspenders" safety. Already the code in the test server is written in a way to ensure serialization. Close() is blocking and returns when the test server is shut down.

cmd/cogito/main_test.go
```diff
@@ -62,14 +70,45 @@ func TestRunPutSmokeIntegrationSuccess(t *testing.T) {
...
+	assert.Equal(t, ghReq.State, string(wantState), "", ghReq.State, string(wantState))
+	assert.Equal(t, path.Base(URL.Path), wantGitRef)
 }
 
-func TestRunSmokeFailure(t *testing.T) {
+func TestRunPutSuccessIntegration(t *testing.T) {
+	if testing.Short() {
```
marco-m-pix4d, line 85: This is set when the `-short` flag is passed to `go test`.
Compare the Task targets `test:unit` and `test:all` to see how it is used.
File CONTRIBUTING also explains this.

cmd/cogito/main_test.go
```diff
@@ -62,14 +70,45 @@ func TestRunPutSmokeIntegrationSuccess(t *testing.T) {
...
+		t.Skip("skipping integration test")
+	}
+
+	cfg := testhelp.SkipTestIfNoEnvVars(t)
```
marco-m-pix4d, line 89: Fill cfg with environment variables containing the secrets. See Task target `test:env` for the interaction with `gopass`.

cogito/ghcommitsink.go
```diff
@@ -0,0 +1,63 @@
...
+// The states allowed by cogito are more than the states allowed by the GitHub Commit
+// status API. Adapt accordingly.
+func ghAdaptState(state BuildState) string {
+	if state == StateAbort {
+		return string(StateError)
+	}
+	return string(state)
+}
+
+// ghMakeContext returns the "context" parameter of the GitHub Commit Status API, based
```
marco-m-pix4d, line 50: See the README for details.

cogito/ghcommitsink_test.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+func TestSinkGitHubCommitStatusSendSuccess(t *testing.T) {
+	wantGitRef := "deadbeefdeadbeef"
+	wantState := cogito.StatePending
+	jobName := "the-job"
+	wantContext := jobName
+	var ghReq github.AddRequest
+	var URL *url.URL
+	ts := testhelp.GhCommitStatusTestServer(&ghReq, &URL)
```
marco-m-pix4d, line 24: Another usage of the mock http server

cogito/ghcommitsink_test.go
```diff
@@ -0,0 +1,63 @@
...
+	assert.Equal(t, path.Base(URL.Path), wantGitRef)
+	assert.Equal(t, ghReq.State, string(wantState))
+	assert.Equal(t, ghReq.Context, wantContext)
+}
+
+func TestSinkGitHubCommitStatusSendFailure(t *testing.T) {
+	ts := httptest.NewServer(
+		http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			w.WriteHeader(http.StatusTeapot)
+		}))
```
marco-m-pix4d, line 48: An always failing HTTP server, with code 418.

github/commitstatus_test.go
```diff
@@ -177,7 +177,7 @@ OAuth: X-Accepted-Oauth-Scopes: [], X-Oauth-Scopes: []`,
 			name: "non existing repo: Not Found",
 			repo: "non-existing-really",
 			wantErr: `Failed to add state "success" for commit 32e4b4f: 404 Not Found
-Body: {"message":"Not Found","documentation_url":"https://docs.github.com/rest/reference/repos#create-a-commit-status"}
+Body: {"message":"Not Found","documentation_url":"https://docs.github.com/rest/commits/statuses#create-a-commit-status"}
```
marco-m-pix4d, line 180: Example of flaky test because too tight and dependent on upstream.
Maybe we should do a partial string match instead...

resource/chatadapter.go
```diff
@@ -25,7 +25,8 @@ func sendToChat(
 	team := env.Get("BUILD_TEAM_NAME")
 	buildN := env.Get("BUILD_NAME")
 	instanceVars := env.Get("BUILD_PIPELINE_INSTANCE_VARS")
-	buildURL := concourseBuildURL(atc, team, pipeline, job, buildN, instanceVars)
```
marco-m-pix4d, line 28: This is the chatadapter, not yet integrated. This change is only to be able to compile and get rid of GH static analysis.

testhelp/testhelper.go
```diff
@@ -35,24 +34,26 @@ func IdentityRenamer(name string) string {
...
+//     file will be renamed accordingly.
 //
 // It will fail if the dst directory doesn't exist.
 //
 // For example, if src directory is `foo`:
 // foo
 // └── dot.git
-//     └── config
+//
+//	└── config
```
marco-m-pix4d, line 49: This is my over-zealous editor :-(

- [ ] fix zealous and wrong editor

testhelp/testhelper.go
```diff
@@ -70,7 +71,7 @@ func CopyDir(dst string, src string, dirRenamer Renamer, templatedata TemplateDa
 		return fmt.Errorf("making src dir: %w", err)
 	}
 
-	srcEntries, err := ioutil.ReadDir(src)
+	srcEntries, err := os.ReadDir(src)
```
marco-m-pix4d, line 74: Package ioutils is deprecated

testhelp/testserver.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+)
+
+// GhCommitStatusTestServer returns a very basic HTTP test server (mock) that speaks
```
marco-m-pix4d, line 11: I like how this turned out :-)

github/commitstatus.go
```diff
@@ -122,7 +127,7 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 	OAuthInfo := fmt.Sprintf("X-Accepted-Oauth-Scopes: %v, X-Oauth-Scopes: %v",
 		XAcceptedOauthScope, XOauthScopes)
 
-	respBody, _ := ioutil.ReadAll(resp.Body)
+	respBody, _ := io.ReadAll(resp.Body)
```
Esysc, line 130: What is the difference here? I'm quite confused as it seems there are equivalent methods to do the same.

Esysc: > What is the difference here? I'm quite confused as it seems there are equivalent methods to do the same.

Ok, I got the answer, you comment was not shown while reviewing

cogito/put.go
```diff
@@ -160,9 +160,13 @@ func (pu *ProdPutter) ProcessInputDir() error {
...
 func (pu *ProdPutter) Sinks() []Sinker {
-	pu.log.Info("prodPutter: Sinks")
 	return []Sinker{
-		GitHubCommitStatusSink{Putter: pu},
+		GitHubCommitStatusSink{
+			Log:     pu.log,
+			GhAPI:   pu.ghAPI,
+			GitRef:  pu.gitRef,
+			Request: pu.Request,
+		},
```
marco-m-pix4d, line 169: @odormond you remember your comment in previous PR on how to make the sink generic?
This how it is done: the Putter initializes it, setting the fields it needs.
This allows to have, in Put(), a simple and modular `sink.Send() error` without parameters.

Having sink.LoadConfiguration() as you were suggesting would, as I understand it, have two possible implementations:

1. A parameterless LoadConfiguration like Send: `sink.LoadConfiguration() error`. This would require the same mechanism used here in line 164, so no need for LoadConfiguration at all.
2. A LoadConfiguration with parameters. But in this case, how to keep it generic (modular)? Only option would be a io.Reader parameter, and pass to it, _again_, the JSON text to be parsed :-(

I actually though about these options, and this is why, after many different approaches, I selected what we have now.

To be complete, starting with Go 1.18, we could use generics, and so we could pass to sink.Send (or to sink.LoadConfiguration) a type parameter that depends on the implementation (cogito would pass a cogito.PutRequest, another resource foo would pass a foo.PutRequest ...). I tried also this one, but at the end the (slightly) added complexity did not seem justified to me...

(it took a while to come out with the current design, it was not evident!)

odormond: Thanks for the explanation! I did not consider that the static typing was indeed a problem for a more modular approach.

odormond (approved):

Esysc (approved):

PR #91, opened 2022-08-10
Remove of course 7: put part 4: add sink Google chat

marco-m-pix4d:
add sink google chat; remove remnants of the ofcourse pkg

This is finally back to feature parity with the master branch, with added a lot more
test coverage and overall cleaner, type safe and extensible design.

Best reviewed commit-per-commit; the commit message gives context.

Test coverage of the cogito package is at 100% 💪 

PCI-2587

TODO

- [x] move `statesToNotifyChat` inline


cmd/cogito/main_test.go
```diff
@@ -52,60 +55,77 @@ func TestRunGetSuccess(t *testing.T) {
...
 
 	err := run(in, &out, &logOut, []string{"out", inputDir})
 
-	assert.NilError(t, err, "\nout: %s\nlogOut: %s", out.String(), logOut.String())
+	assert.NilError(t, err, "\nout:\n%s\nlogOut:\n%s", out.String(), logOut.String())
+	assert.Assert(t, cmp.Contains(logOut.String(),
+		"cogito.put.ghCommitStatus: commit status posted successfully"))
+	assert.Assert(t, cmp.Contains(logOut.String(),
+		"cogito.put.gChat: state posted successfully to chat"))
+	// t.Logf("\nlogOut:\n%s", logOut.String())
```
odormond, line 128: Leftover printf debugging?

cogito/gchatsink.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/Pix4D/cogito/googlechat"
+	"github.com/hashicorp/go-hclog"
+)
+
+// TODO FIND A BETTER PLACE?
```
odormond, line 12: No strong opinion on that one, but given it's private and used in a single place, I think I would have just inlined it in `shouldSendToChat`.

Should it become a tunable, it can be moved to were the parsing takes place as it would become the default.

odormond (approved):

cogito/gchatsink.go
marco-m-pix4d: Yes, exactly for both comments. In this PR I will just inline, and in the next ticket I will move close to the parsing.

marco-m-pix4d: Done.

cmd/cogito/main_test.go
marco-m-pix4d: Done.

Esysc (approved):
LGTM!

PR #93, opened 2022-08-15
Prepare override of chat states

marco-m-pix4d:
This PR contains two somewhat unrelated cleanups: (1) document some more release workarounds (2) introduce custom JSON parsing for `build_state` to keep type safety and auto-validate.

- doc: more workarounds for concourse bug not picking up latest
- BuildState: add custom unmarshaller for validation
    This also means that we can remove the explicit call to BuildState.Validate() in
    ProdPutter.LoadConfiguration(), since it will be done by the JSON parser when
    it encounters the BuildState type as a field of struct PutRequest.

PCI-2586

### DIFF
--- a/cogito/protocol.go
+++ b/cogito/protocol.go
@@ -4,6 +4,7 @@
 package cogito
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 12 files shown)


cogito/protocol.go
```diff
@@ -199,13 +200,24 @@ const (
 
 const KeyState = "state"
 
-// Validate checks whether the build state, parsed from JSON, is valid.
-func (bs BuildState) Validate() error {
-	switch bs {
+func (bs *BuildState) UnmarshalJSON(data []byte) error {
```
marco-m-pix4d, line 203: To better understand what is going on here, see for example https://pkg.go.dev/encoding/json@go1.19#example-package-CustomMarshalJSON

Esysc (approved):

odormond (approved):

PR #94, opened 2022-08-16
Safe parse URL to not leak password in case of error

iAmoric:
Parses a given URL and only returns the error to avoid potential password leaks

Part of PCI-2625

- [x] Merge **only** after fixing tests

### DIFF
--- a/cogito/put.go
+++ b/cogito/put.go
@@ -2,6 +2,7 @@ package cogito
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/url"
@@ -274,14 +275,32 @@ type gitURL struct {
 	Repo  string
 }
 
+// safeParseUrl parse the given URL and return an error if the URL is not well formated.
+// this avoid to eventually leak passwords in case of error
+//
+// From https://github.com/golang/go/issues/53993
+func safeParse(rawURL string) (*url.URL, error) {
+	parsedUrl, err := url.Parse(rawURL)
+	if err != nil {
+		// check if the returned error is actually of type url.Error
+		if uerr := new(url.Error); errors.As(err, &uerr) {
+			// return only the Err and not the URL to not leak potential passwords
+			return nil, uerr.Err
+		}
+		// default error
+		return nil, errors.New("invalid URL")
+	}
+	return parsedUrl, nil
+}
+
 // parseGitPseudoURL attempts to parse rawURL as a git remote URL compatible with the
 // Github naming conventions.
 //
 // It supports the following types of git pseudo URLs:
 //   - ssh:   git@github.com:Pix4D/cogito.git; will be rewritten to the valid URL
 //     ssh://git@github.com/Pix4D/cogito.git
-//   - https: https://github.com/Pix4D/cogito.git
-//   - http:  http://github.com/Pix4D/cogito.git
+//   - https: https://github.com/Pix4D/cogito.git or https://username:password@github.com/Pix4D/cogito.git
+//   - http: http://github.com/Pix4D/cogito.git or http://username:password@github.com/Pix4D/cogito.git
 func parseGitPseudoURL(rawURL string) (gitURL, error) {
 	workURL := rawURL
 	// If ssh pseudo URL, we need to massage the rawURL ourselves :-(
@@ -294,7 +313,7 @@ func parseGitPseudoURL(rawURL string) (gitURL, error) {
 		workURL = "ssh://" + strings.Replace(workURL, ":", "/", 1)
 	}
 
-	anyUrl, err := url.Parse(workURL)
+	anyUrl, err := safeParse(workURL)
 	if err != nil {
 		return gitURL{}, err
 	}
--- a/cogito/put.go
+++ b/cogito/put.go
@@ -282,7 +282,8 @@ type gitURL struct {
 func safeUrlParse(rawURL string) (*url.URL, error) {
 	parsedUrl, err := url.Parse(rawURL)
 	if err != nil {
-		if uerr := new(url.Error); errors.As(err, &uerr) {
+		var uerr *url.Error
+		if errors.As(err, &uerr) {
 			// url.Parse returns a wrapped error that contains also the URL. Instead, we return only the error.
 			return nil, uerr.Err
 		}
--- a/cogito/put.go
+++ b/cogito/put.go
@@ -284,7 +284,8 @@ func safeUrlParse(rawURL string) (*url.URL, error) {
 	if err != nil {
 		var uerr *url.Error
 		if errors.As(err, &uerr) {
-			// url.Parse returns a wrapped error that contains also the URL. Instead, we return only the error.
+			// url.Parse returns a wrapped error that contains also the URL.
+			// Instead, we return only the error.
 			return nil, uerr.Err
 		}
 		return nil, errors.New("invalid URL")
--- a/cogito/put.go
+++ b/cogito/put.go
@@ -275,19 +275,17 @@ type gitURL struct {
 	Repo  string
 }
 
-// safeParseUrl parse the given URL and return an error if the URL is not well formated.
-// this avoid to eventually leak passwords in case of error
+// safeUrlParse wraps [url.Parse] and returns only the error and not the URL to avoid leaking
+// passwords of the form http://user:password@example.com
 //
 // From https://github.com/golang/go/issues/53993
-func safeParse(rawURL string) (*url.URL, error) {
+func safeUrlParse(rawURL string) (*url.URL, error) {
 	parsedUrl, err := url.Parse(rawURL)
 	if err != nil {
-		// check if the returned error is actually of type url.Error
 		if uerr := new(url.Error); errors.As(err, &uerr) {
-			// return only the Err and not the URL to not leak potential passwords
+			// url.Parse returns a wrapped error that contains also the URL. Instead, we return only the error.
 			return nil, uerr.Err
 		}
-		// default error
 		return nil, errors.New("invalid URL")
 	}
 	return parsedUrl, nil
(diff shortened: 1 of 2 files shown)


cogito/put.go
```diff
@@ -277,14 +278,32 @@ type gitURL struct {
 	Repo  string
 }
 
+// safeParseUrl parse the given URL and return an error if the URL is not well formated.
+// this avoid to eventually leak passwords in case of error
+//
+// From https://github.com/golang/go/issues/53993
+func safeParse(rawURL string) (*url.URL, error) {
```
iAmoric, line 285: Due to the new URL parsing, in case of error, we only show the error and not the faulty URL. This means we can lose (important) debug information.
This is also causing tests `invalid HTTPS URL` and `invalid HTTP URL` to fail because they currently expect the URL in the response

cogito/put_private_test.go
```diff
@@ -264,11 +292,13 @@ func TestParseGitPseudoURLFailure(t *testing.T) {
 			wantErr: "invalid git SSH URL git@github.com/Pix4D/cogito.git: want exactly one ':'",
 		},
 		{
+			// currently failing due to the new safeParse method returning only the error
```
iAmoric, line 295:
```
    --- FAIL: TestParseGitPseudoURLFailure/invalid_HTTPS_URL (0.00s)
        put_private_test.go:327: assertion failed: expected error "parse \"https://github.com:Pix4D/cogito.git\": invalid port \":Pix4D\" after host", got "invalid port \":Pix4D\" after host"
    --- FAIL: TestParseGitPseudoURLFailure/invalid_HTTP_URL (0.00s)
        put_private_test.go:327: assertion failed: expected error "parse \"http://github.com:Pix4D/cogito.git\": invalid port \":Pix4D\" after host", got "invalid port \":Pix4D\" after host"
```

Esysc: I think that having only the error should be enough, in this case `invalid port` is useful. The URL is in the pipeline configuration if we really need it.

cogito/put.go
```diff
@@ -277,14 +278,32 @@ type gitURL struct {
...
+	parsedUrl, err := url.Parse(rawURL)
+	if err != nil {
+		// check if the returned error is actually of type url.Error
+		if uerr := new(url.Error); errors.As(err, &uerr) {
+			// return only the Err and not the URL to not leak potential passwords
+			return nil, uerr.Err
+		}
+		// default error
+		return nil, errors.New("invalid URL")
```
Esysc, line 294: When we reach this point, are we 100% sure that the URL was invalid?

cogito/put_private_test.go
```diff
@@ -264,11 +292,13 @@ func TestParseGitPseudoURLFailure(t *testing.T) {
...
 			name:    "invalid HTTPS URL",
 			inURL:   "https://github.com:Pix4D/cogito.git",
 			wantErr: `parse "https://github.com:Pix4D/cogito.git": invalid port ":Pix4D" after host`,
```
odormond, line 298: Why not just fix the test?
```suggestion
			wantErr: `invalid port ":Pix4D" after host`,
```

cogito/put.go
```diff
@@ -277,14 +278,32 @@ type gitURL struct {
...
+// From https://github.com/golang/go/issues/53993
+func safeParse(rawURL string) (*url.URL, error) {
+	parsedUrl, err := url.Parse(rawURL)
+	if err != nil {
+		// check if the returned error is actually of type url.Error
+		if uerr := new(url.Error); errors.As(err, &uerr) {
+			// return only the Err and not the URL to not leak potential passwords
+			return nil, uerr.Err
+		}
+		// default error
```
marco-m-pix4d, line 293: remove comment

cogito/put.go
```diff
@@ -277,14 +278,32 @@ type gitURL struct {
...
 }
 
+// safeParseUrl parse the given URL and return an error if the URL is not well formated.
+// this avoid to eventually leak passwords in case of error
+//
+// From https://github.com/golang/go/issues/53993
+func safeParse(rawURL string) (*url.URL, error) {
+	parsedUrl, err := url.Parse(rawURL)
+	if err != nil {
+		// check if the returned error is actually of type url.Error
```
marco-m-pix4d, line 288: Remove comment.
I understand it is new to you, but this is part of the Go syntax, so no comment needed.

cogito/put.go
```diff
@@ -277,14 +278,32 @@ type gitURL struct {
...
+		if uerr := new(url.Error); errors.As(err, &uerr) {
+			// return only the Err and not the URL to not leak potential passwords
```
marco-m-pix4d, line 290: // Return only the error and not the URL to avoid leaking passwords of the form http://user:password@example.com

cogito/put_private_test.go
iAmoric: Actually because for now I don't really know if I should either fix the test to remove the URL, or adapt the method to keep some debug information

cogito/put.go
marco-m-pix4d: Mhh. Actually this should be in the function comment, here it is enough

```
// Return only the error and not the URL (as url.Parse does).
```

cogito/put.go
```diff
@@ -277,14 +278,32 @@ type gitURL struct {
 	Repo  string
 }
 
+// safeParseUrl parse the given URL and return an error if the URL is not well formated.
+// this avoid to eventually leak passwords in case of error
```
marco-m-pix4d, line 282:
```
// safeUrlParse wraps [url.Parse] and returns only the error and not the URL to avoid leaking
// passwords of the form http://user:password@example.com
```

cogito/put.go
marco-m-pix4d, line 285: Suggest to rename to `safeUrlParse` to mimic the name of the wrapped function, `url.Parse`.

cogito/put.go
iAmoric: This means that the URL parsing went wrong and the function `url.Parse` returned a non-nil error.
Maybe I could change the error message with `unable to parse URL` instead of `Invalid URL`?

iAmoric:
@Pix4D/integration the debate here is to know if we accept to lose some debug information to avoid leaking passwords. 
I'm ready to receive your opinions and adapt the PR accordingly :)

cogito/put.go
```diff
@@ -277,14 +278,30 @@ type gitURL struct {
...
 // parseGitPseudoURL attempts to parse rawURL as a git remote URL compatible with the
 // Github naming conventions.
 //
 // It supports the following types of git pseudo URLs:
 //   - ssh:   git@github.com:Pix4D/cogito.git; will be rewritten to the valid URL
 //     ssh://git@github.com/Pix4D/cogito.git
-//   - https: https://github.com/Pix4D/cogito.git
-//   - http:  http://github.com/Pix4D/cogito.git
+//   - https: https://github.com/Pix4D/cogito.git or https://username:password@github.com/Pix4D/cogito.git
+//   - http: http://github.com/Pix4D/cogito.git or http://username:password@github.com/Pix4D/cogito.git
```
marco-m-pix4d, line 304: Could you please put each example on its own line? The versions with the u:p are different enough to be made more visible.

cogito/put_private_test.go
```diff
@@ -228,6 +228,34 @@ func TestParseGitPseudoURLSuccess(t *testing.T) {
...
+					User:   url.UserPassword("username", "password"),
+					Host:   "github.com",
+					Path:   "/Pix4D/cogito.git",
+				},
+				Owner: "Pix4D",
+				Repo:  "cogito",
+			},
+		},
+		{
+			name:  "valid HTTP URL with username:password",
```
marco-m-pix4d, line 246: s/HTTP/HTTPS/

cogito/put.go
iAmoric: Fixed in [69d2474](https://github.com/Pix4D/cogito/pull/94/commits/69d24744e01f19e6047d6393bfb78160aa816d76)

Esysc:
> @Pix4D/integration the debate here is to know if we accept to lose some debug information to avoid leaking passwords. I'm ready to receive your opinions and adapt the PR accordingly :)

I think that we can get rid of the URL in the message, it's something known in any case.

cogito/put.go
marco-m-pix4d: > When we reach this point, are we 100% sure that the URL was invalid?

Actually this point (line 292) is currently unreachable, since url.Parse (for the time being) always returns type url.Error.

But to answer the question: yes, we are 100% sure, because it is still under the error condition of url.Parse (see line 287 where the branch begins).

cogito/put.go
```diff
@@ -277,14 +278,32 @@ type gitURL struct {
...
 }
 
+// safeUrlParse wraps [url.Parse] and returns only the error and not the URL to avoid leaking
+// passwords of the form http://user:password@example.com
+//
+// From https://github.com/golang/go/issues/53993
+func safeUrlParse(rawURL string) (*url.URL, error) {
+	parsedUrl, err := url.Parse(rawURL)
+	if err != nil {
+		if uerr := new(url.Error); errors.As(err, &uerr) {
```
marco-m-pix4d, line 288: I realized that this code, although correct, can be confusing.
The idiomatic way to use errors.As is
```
		var uerr *url.Error
		if errors.As(err, &uerr) {
```
@iAmoric  Could you please update?

See https://go.dev/blog/go1.13-errors

marco-m-pix4d (approved):
I tried a few things but I think that what you propose here is the best we can have.

Could you please
- have a look at my lasts comments
- ensure that the line length is no more than 90, as this is the limit for this project
- squash what can be squashed

thanks! Good work!

cogito/put.go
iAmoric: Fixed in [c544b0c](https://github.com/Pix4D/cogito/pull/94/commits/c544b0c86c4f8398ac6ce008fdacfa21a0ef9f0f)

Esysc (approved):

PR #96, opened 2022-08-17
better chat message

marco-m-pix4d:
Now the job and the commit are clickable links.

Before

cogito-before

After
note: compared to the screenshot, I removed the team, following @odormond comment.

cogito-after


odormond (approved):

Esysc (approved):

Esysc:
Great the commit in the message, and also the reduced noise of build URL.

iAmoric (approved):

PR #99, opened 2022-08-29
add a generic set data structure

marco-m-pix4d:
All the ones I could find where clumsy to use, I was quite surprised.

Hope that Go will soon gain a sets package in the stdlib so that we can get rid
of this one.

I split this from the current branch I am working on because it is self-contained, but to
understand why I implemented only this partial API you have to wait to see the next PR.

I have to say that generics and type constraints are good stuff!

PCI-2561

### DIFF
--- a/sets/sets.go
+++ b/sets/sets.go
@@ -1,0 +1,80 @@
+// Package sets is a minimal implementation of a generic set data structure.
+
+package sets
+
+import (
+	"fmt"
+	"sort"
+
+	"golang.org/x/exp/constraints"
+)
+
+// Set is a minimal set that takes only ordered types: any type that supports the
+// operators < <= >= >.
+type Set[T constraints.Ordered] struct {
+	items map[T]struct{}
+}
+
+// New returns an empty set with capacity size. The capacity will grow and shrink as a
+// stdlib map.
+func New[T constraints.Ordered](size int) *Set[T] {
+	return &Set[T]{items: make(map[T]struct{}, size)}
+}
+
+// From returns a set from elements.
+func From[T constraints.Ordered](elements ...T) *Set[T] {
+	s := New[T](len(elements))
+	for _, i := range elements {
+		s.items[i] = struct{}{}
+	}
+	return s
+}
+
+// String returns a string representation of s, ordered. This allows to simply pass a
+// sets.Set as parameter to a function that expects a fmt.Stringer interface and obtain
+// a comparable string.
+func (s *Set[T]) String() string {
+	return fmt.Sprint(s.OrderedList())
+}
+
+func (s *Set[T]) Size() int {
+	return len(s.items)
+}
+
+// OrderedList returns a slice of the elements of s, ordered.
+// TODO This can probably be replaced in Go 1.20 when a generics slice packages reaches
+// the stdlib.
+func (s *Set[T]) OrderedList() []T {
+	elements := make([]T, 0, len(s.items))
+	for e := range s.items {
+		elements = append(elements, e)
+	}
+	sort.Slice(elements, func(i, j int) bool {
+		return elements[i] < elements[j]
+	})
+	return elements
+}
+
+// Contains returns true if s contains item.
+func (s *Set[T]) Contains(item T) bool {
+	_, found := s.items[item]
+	return found
+}
+
+// Difference returns a set containing the elements of s that are not in x.
+func (s *Set[T]) Difference(x *Set[T]) *Set[T] {
+	result := New[T](max(0, s.Size()-x.Size()))
+	for i := range s.items {
+		if !x.Contains(i) {
+			result.items[i] = struct{}{}
+		}
+	}
+	return result
+}
+
+func max(a, b int) int {
+	if a > b {
+		return a
+	}
+	return b
+}
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/hashicorp/go-hclog v1.2.2
 	github.com/imdario/mergo v0.3.13
 	github.com/sasbury/mini v0.0.0-20181226232755-dc74af49394b
+	golang.org/x/exp v0.0.0-20220827204233-334a2380cb91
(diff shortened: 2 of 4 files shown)


sets/sets.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Set is a minimal set that takes only ordered types: any type that supports the
+// operators < <= >= >.
+type Set[T constraints.Ordered] struct {
+	items map[T]struct{}
+}
+
+// New returns an empty set with capacity size. The capacity will grow and shrink as a
+// stdlib map.
+func New[T constraints.Ordered](size int) *Set[T] {
```
marco-m-pix4d, line 20: To understand better, see https://go.dev/blog/intro-generics

sets/sets.go
```diff
@@ -0,0 +1,80 @@
...
+	return &Set[T]{items: make(map[T]struct{}, size)}
+}
+
+// From returns a set from elements.
+func From[T constraints.Ordered](elements ...T) *Set[T] {
+	s := New[T](len(elements))
+	for _, i := range elements {
+		s.items[i] = struct{}{}
```
marco-m-pix4d, line 28: The underlying data structure is the workhorse Go map.
The type `struct{}{}` is special and takes 0 bytes and is the idiomatic way to implement a set in Go.

sets/sets.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	return s
+}
+
+// String returns a string representation of s, ordered. This allows to simply pass a
+// sets.Set as parameter to a function that expects a fmt.Stringer interface and obtain
+// a comparable string.
+func (s *Set[T]) String() string {
```
marco-m-pix4d, line 36: See the tests for details if you are rusty about what fmt.Stringer is.

sets/sets.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	sort.Slice(elements, func(i, j int) bool {
+		return elements[i] < elements[j]
+	})
+	return elements
+}
+
+// Contains returns true if s contains item.
+func (s *Set[T]) Contains(item T) bool {
+	_, found := s.items[item]
```
marco-m-pix4d, line 60: Idiomatic map lookup.
It can be confusing, see https://gobyexample.com/maps for details; notice that the lookup can be done assigning to one or to two variables (special case).

Esysc (approved):

sets/sets.go
```diff
@@ -0,0 +1,80 @@
...
+package sets
+
+import (
+	"fmt"
+	"sort"
+
+	"golang.org/x/exp/constraints"
+)
+
+// Set is a minimal set that takes only ordered types: any type that supports the
```
odormond, line 12: Wouldn't equality be enough as a requirement?

sets/sets.go
```diff
@@ -0,0 +1,80 @@
...
+// From returns a set from elements.
+func From[T constraints.Ordered](elements ...T) *Set[T] {
+	s := New[T](len(elements))
+	for _, i := range elements {
+		s.items[i] = struct{}{}
+	}
+	return s
+}
+
+// String returns a string representation of s, ordered. This allows to simply pass a
```
odormond, line 33: Ah. That's why you want more than equality.

sets/sets.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Contains returns true if s contains item.
+func (s *Set[T]) Contains(item T) bool {
+	_, found := s.items[item]
+	return found
+}
+
+// Difference returns a set containing the elements of s that are not in x.
+func (s *Set[T]) Difference(x *Set[T]) *Set[T] {
+	result := New[T](max(0, s.Size()-x.Size()))
```
odormond, line 66: Err... if `x` and `s` are disjoint the result will be equal to `s` so the actual max size it can reach is the size of `s` no matter what's the size of `x` is.

odormond (approved):
I'm pre-approving but the difference pre-sizing computation should be dropped, IMO. Simply set it to the size of `s`.

sets/sets.go
marco-m-pix4d: The reason, as you discovered below in https://github.com/Pix4D/cogito/pull/99/files#r959248227, is to be able to have a repeatable string representation.

There are other ways to skin this cat for sure (and I considered some), but this is the absolute minimum that works for the current Cogito use case.

sets/sets.go
marco-m-pix4d: Correct. We are allocating memory, and in this case we must allocate it before knowing how much we really need.

This errs on the side of allocating as little as possible, which is still better than not allocating anything. The map will then grow automatically with the inserts.

We could allocate the maximum size (`s.Size()`) and then release what is not used, but Go doesn't currently have a way to shrink memory allocated to a map (except playing GC tricks), see https://github.com/golang/go/issues/54766.

Last, we could just not preallocate anything. But it would have no positive impact compared to the current code, actually it would incur more allocations.

Or am I missing something?

odormond: If x is larger than s, you're allocating 0.
If x is smaller than s, you're still likely to be under-allocating unless x is totally included in s which shouldn't be the general case.
So you're almost always under-allocating. Is it worth it? If you really want to avoid the growth, then pre-allocate s. For the current use case of having N input directories where N is small (<10), the waste is probably much better then the reallocation+move.

marco-m-pix4d: > So you're almost always under-allocating. Is it worth it?

That is the point. It is always worth it, in the sense that it cannot do harm and sometimes it does good.
For sure, for the sizes we are talking about for cogito usage, as you point out, the difference is unmeasurable.
Still, there is no harm, so I just want to be sure I understand your point.

odormond: I just don't think it's worth it.

I don't know how Go handles the growth but the usual way is to double the current allocation (when still under some limit) whenever you need more room and so you end up with over-allocation anyway in most cases.

marco-m-pix4d: @odormond I understand your position. We will have to agree to disagree :-)

PR #100, opened 2022-08-30
Add custom chat message (string and file)

marco-m-pix4d:
This should close the current tickets PCI-2561 (cogito: independent chat send) and PCI-2649 (cogito: custom chat message).

Configuration changes for the `put` step:

- `chat_message`: Custom chat message; overrides the build summary. Default: empty.
- `chat_message_file`: Path to file containing a custom chat message; overrides the build summary. Appended to `chat_message`. Default: empty.
- `chat_message_append`: (one of: true, false). If true, append the default build summary to the custom `chat_message` and/or `chat_message_file`. Default: false.

Example:

```
  - put: cogito
    inputs: [target-repo.git, messagedir]
    params:
      state: success
      chat_message: "Custom chat message. Message file and build summary appended"
      chat_message_file: "messagedir/message.txt"
      chat_message_append: true
```

See also the new acceptance tests pipeline `pipelines/cogito-acceptance.yml` and how it is triggered from Task.

TODO

- [x] CONTRIBUTING: mention new gopass "secret" cogito/concourse_url
- [x] test with trailing whitespace: explicitly explain why we have the whitespace
- [x] Taskfile: make fly commands uniform (all long or all short)
- [x] detecting which directory is which: try to simplify the logic following Olivier's suggestion
- [x] README: explain that the put.inputs can be two and mention this in the put params for visibility (discussion:  https://github.com/Pix4D/cogito/pull/100#discussion_r959674410)


cogito/gchatsink.go
```diff
@@ -3,6 +3,7 @@ package cogito
 import (
 	"context"
 	"fmt"
+	"io/fs"
```
marco-m-pix4d, line 6: Like afero FS, but in the Go stdlib (makes sense only for read-only FS, like the one used by Cogito)

cogito/gchatsink_private_test.go
```diff
@@ -1,8 +1,11 @@
 package cogito
 
 import (
+	"strings"
 	"testing"
+	"testing/fstest"
```
marco-m-pix4d, line 6: The test counterpart of `io/fs`, courtesy of the stdlib.

cogito/gchatsink_test.go
```diff
@@ -127,3 +128,19 @@ func TestSinkGoogleChatSendFailure(t *testing.T) {
...
 }
+
+func TestSinkGoogleChatSendInputFailure(t *testing.T) {
+	request := basePutRequest
+	request.Params.ChatMessageFile = "foo/msg.txt"
+	request.Source.GChatWebHook = "dummy-url"
+	assert.NilError(t, request.Source.Validate())
+	sink := cogito.GoogleChatSink{
+		Log:      hclog.NewNullLogger(),
+		InputDir: fstest.MapFS{"bar/msg.txt": {Data: []byte("from-custom-file")}},
```
marco-m-pix4d, line 139: the vfs in action

cogito/put.go
```diff
@@ -136,21 +138,67 @@ func (putter *ProdPutter) LoadConfiguration(in io.Reader, args []string) error {
...
+	// If on the other hand it contains two, one should be the git repo (still nameless)
+	// and the other should be the directory containing the chat_message_file, which is
+	// named by the first element of the path in "chat_message_file".
+	// This allows (although clumsily) to distinguish which is which.
+	// This complexity has historical reasons to preserve backwards compatibility
+	// (the nameless git repo).
+	//
+	// Somehow independent is the reason why we enforce the count of directories to be
+	// max 2: this is to avoid the default Concourse behavior of streaming _all_ the
+	// volumes "just in case".
```
marco-m-pix4d, line 153: This explains why the code is convoluted. I am happy if we find a simpler approach.

pipelines/cogito-acceptance.yml
```diff
@@ -0,0 +1,68 @@
+# NOTE
```
marco-m-pix4d, line 1: See the note here

Taskfile.yml
```diff
@@ -117,6 +119,43 @@ tasks:
         # stderr for logging.
         sh: echo '{{.INPUT}}' | ./bin/check 2>&1 1>/dev/null
 
+  test:acceptance:
```
iAmoric, line 122: Maybe you can add a `precondition` step to automatically run `fly login` when we need to run acceptance tests (and maybe also a defer to fly logout? )

README.md
```diff
@@ -198,10 +198,16 @@ If the `source` block has the optional key `gchat_webhook`, then it will also se
 
 - `state`: The state to set. One of `error`, `failure`, `pending`, `success`, `abort`. Subject to the mapping explained in section [Effects](#effects).
 
-## Optional params
+## Optional params for GitHub commit status
 
 - `context`: The value of the non-prefix part of the GitHub Commit status API "context" (see section [Effects on GitHub](#effects-on-github)). Default: the job name. See also the optional `context_prefix` in the [source configuration](#source-configuration).
+
+## Optional params for chat
```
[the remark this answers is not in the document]
marco-m-pix4d: It is generic on purpose, because we might be adding other chat backends.

README.md
```diff
@@ -198,10 +198,16 @@ If the `source` block has the optional key `gchat_webhook`, then it will also se
...
+
 - `gchat_webhook`: If present, overrides `source.gchat_webhook`. Default: empty, thus the same as `source.gchat_webhook`. This allows to use the same Cogito resource for multiple chat spaces. 
+- `chat_message`: Custom chat message; overrides the build summary. Default: empty.
+- `chat_message_file`: Path to file containing a custom chat message; overrides the build summary. Appended to `chat_message`. Default: empty.
+- `chat_message_append`: (one of: true, false). If true, append the default build summary to the custom `chat_message` and/or `chat_message_file`. Default: false.
```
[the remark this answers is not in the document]
marco-m-pix4d: Correct.

cogito/gchatsink_private_test.go
```diff
@@ -55,7 +58,118 @@ func TestShouldSendToChatCustomConfig(t *testing.T) {
...
+				}),
+			wantPresent: append(basePresent, "the-custom-message"),
+		},
+		{
+			name: "chat message file",
+			request: testhelp.MergeStructs(
+				baseRequest,
+				PutRequest{Params: PutParams{ChatMessageFile: "registration/msg.txt"}}),
+			inputDir: fstest.MapFS{
+				"registration/msg.txt": {Data: []byte("from-custom-file")},
```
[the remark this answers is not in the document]
marco-m-pix4d: It creates a memory-based file system, because `fstest.MapFS` implements the `io/fs` interface.
If you follow the path to the type of `inputDir` in this test and then to the type of the parameter in the SUT (system under test), `prepareChatMessage()` it should be clearer. In this PR there is also a comment for `prepareChatMessage()`.

cogito/put.go
```diff
@@ -136,21 +138,67 @@ func (putter *ProdPutter) LoadConfiguration(in io.Reader, args []string) error {
...
 
-	// Since we require InputDir to contain only one directory, we assume that this
-	// directory is the git repo.
-	repoDir := filepath.Join(putter.InputDir, inputDirs[0])
-	if err := checkGitRepoDir(
-		repoDir, putter.Request.Source.Owner, putter.Request.Source.Repo); err != nil {
+	// The set has one or two elements. if it exists, remove from the set the message
+	// directory. The remaining one is the git repo.
+	putter.log.Debug("", "inputDirs", inputDirs, "msgDir", msgDir)
+	remaining := inputDirs.Difference(sets.From(msgDir))
```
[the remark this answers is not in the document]
marco-m-pix4d: Yes. The reasoning is explained in the longer comment in this same function, a bit above.

Taskfile.yml
marco-m-pix4d: Actually at the beginning I had `fly login` as the first step of target `test:acceptance`. Then I found it annoying, because it keeps popping up the web browser also if it is already logged in.

Currently you are supposed to call once `task test:acceptance:fly-login` and then call `task test:acceptance` as many times as you need. I found this works well when developing the acceptance tests or a new feature. I think it is the less-worse trade-off...

cogito/put.go
odormond: Personally, I would have done something like this (pseudo code)
```
inputDirs = collectInputDirs(putter.InputDir)
if params.ChatMessageFile != "" {
  msgDir, _ = path.Split(params.ChatMessageFile)
  msgDir = strings.TrimSuffix(msgDir, "/")
  if msgDir == "" {
    return fmt.Errorf("blabla wrong format")
  }
  _, found = inputDirs.pop(msgDir)
  if !found {
    return fmt.Errorf("blabla not found")
  }
}
if inputDirs.size() != 1 {
  return fmt.Errorf("blabla missing git repo")
}
```
But I'm a bit biased by the feature rich sets in Python. Maybe you can still pop from a map in Go?

Even without "pop" you could still avoid checking `ChatMessageFile` twice. If it's there, you first check if you have 2 inputs then check one of them hold the chat message file. Otherwise you just check you've got a single input.

odormond: Thinking a bit more about it, if you Go has lists and trying to remove stuff from it can signal if the element wasn't found, that's enough to implement the same algo and then you wouldn't need a Set type.

Taskfile.yml
```diff
@@ -117,6 +119,43 @@ tasks:
...
+        -y repo-name=$(gopass show cogito/test_repo_name)
+        -y oauth-personal-access-token=$(gopass show cogito/test_oauth_token)
+        -y cogito-tag={{.BRANCH}}
+        -y gchat_webhook=$(gopass show cogito/test_gchat_webhook)
+      - fly -t cogito unpause-pipeline -p cogito-acceptance-{{.BRANCH}}
+
+  test:acceptance:chat-message-file:
+    desc: Run a pipeline job to test chat_message_file
+    cmds:
+      - fly -t cogito tj -j cogito-acceptance-{{.BRANCH}}/chat-message-file.job -w
```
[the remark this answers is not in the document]
marco-m-pix4d: I started using the short versions to keep the commands on one YAML (not too long) line, but I have to agree that now it is not uniform. I also have to agree that the short versions are quite cryptic.

Would you be OK if we use the short versions everywhere? Otherwise, I will come up with something uniform and decently readable.

cogito/put.go
marco-m-pix4d: For sure you can pop from a Go map, even in assembly you can pop :-D

Thanks for the suggestion, let me go through it.

Taskfile.yml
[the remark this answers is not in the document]
odormond: Given we set the line length limit for python code to 99 and here it would just reach 91 I think it's not too long:
```
+>>> len("      - fly -t cogito tj -j cogito-acceptance-{{.BRANCH}}/chat-message-file.job -w")
82
+>>> len("      - fly -t cogito trigger-job -j cogito-acceptance-{{.BRANCH}}/chat-message-file.job -w")
91
```

But we're using short version of the option flags anyway so familiarity with fly is already assumed so pushing further and using short version of the commands should be fine too.

cogito/put.go
```diff
@@ -136,21 +138,67 @@ func (putter *ProdPutter) LoadConfiguration(in io.Reader, args []string) error {
...
+	if params.ChatMessageFile != "" {
+		msgDir, _ = path.Split(params.ChatMessageFile)
+		msgDir = strings.TrimSuffix(msgDir, "/")
+		if msgDir == "" {
+			return fmt.Errorf("chat_message_file: wrong format: have: %s, want: path of the form: <dir>/<file>",
+				params.ChatMessageFile)
+		}
+	}
+
+	collected, err := collectInputDirs(putter.InputDir)
```
Esysc, line 168: Could this block be placed before L159? it'd permit to check earlier the conditions and consequently less `if/else` if I'm not wrong

pipelines/cogito-acceptance.yml
```diff
@@ -0,0 +1,68 @@
...
+
+jobs:
+
+  - name: chat-message-append.job
+    plan:
+      - get: target-repo.git
+      - put: cogito
+        inputs: [target-repo.git]
+        params:
+          state: success
```
aliculPix4D, line 54: So this also pushes the status to github?

Just for my understanding: let say that this put step fails, but after it managed to push success status to github i.e while posting msg to chat. What would happen? In other pipelines we would have on_failure step, so it would post again the status and default build msg, hence overwriting the status from this put step?

README.md
```diff
@@ -198,10 +198,16 @@ If the `source` block has the optional key `gchat_webhook`, then it will also se
 
 - `state`: The state to set. One of `error`, `failure`, `pending`, `success`, `abort`. Subject to the mapping explained in section [Effects](#effects).
 
-## Optional params
+## Optional params for GitHub commit status
```
aliculPix4D, line 201: I think you should also update the note here: https://github.com/Pix4D/cogito#note

> `The put step requires one and only one ["put inputs"](https://concourse-ci.org/put-step.html#put-step-inputs) to be specified, otherwise it will error out. For example:`

pipelines/cogito-acceptance.yml
marco-m-pix4d: Good question.

First, I want to make it clear that I don't have strong opinions on the current behavior. We can change it if we think there is a better way.

Before answering your question, some context. The "problem" is two-fold, and I think that I mentioned it in the design document for this ticket:

1. A `put` on a Concourse resource doesn't know _where_ is running: it could be running at the end of a job via an handler or it could be running in the middle of a job as an explicit `put` after a task. It doesn't even know the build state (see https://github.com/concourse/concourse/issues/8541).
2. Currently we require to always specify the `state`, also when arguably it doesn't make a lot of sense (like in this case).

To answer the question, the sequence should be the following:

a. current job marked as success in github
b. posting message to chat fails, cogito exits 1, concourse considers the job errored (color: orange)
c. if there is an handler, cogito runs again and (if everything goes well) sets the same job as errored
d. if there is no handler, the job remains forever as success on github

Note that something similar can already happen:

1. job starts, cogito sets github state as pending
2. a task exits 1
3. concourse runs the on_failure hook
4. cogito runs and for any reasons it fails to update the state
5. the job remains forever in state pending on github

This makes me realize that we should at the very minimum explain in the cogito readme (we cannot enforce) that a `put` in the middle of a job should have only state `pending`, because in case of subsequent malfunctioning it will be "less wrong" than a `success`, so thanks!

Another somewhat related change I am considering is to do like the hangouts resource, that it always exits 0 also on error, and sets the metadata fields:
```
status	Failed                                <== as resource output metadata
error	Missing 'webhook_url' in source       <== as resource output metadata
```
which might be better?

README.md
aliculPix4D: Also, this makes me wonder:
Should the input be added to required for put step (even now, before your PR)
With current `cogito` if one does not set `inputs: [the-repo]` the resource will fail and this should be added under required for put step. I now that this is not a params, but it should not be under note. 

Also, as I wrote above, note is not up to date with this PR anymore.

pipelines/cogito-acceptance.yml
aliculPix4D: > posting message to chat fails, cogito exits 1, concourse considers the job errored (color: orange)

why errored? Errored is if volume disappears or some other kind of Concourse related issue. This will be red and marked as failure?
Or am I wrong?

README.md
marco-m-pix4d: You are right, I will update the README.

> With current cogito if one does not set inputs: [the-repo] the resource will fail and this should be added under required for put step. I now that this is not a params, but it should not be under note.

I see what you mean. Somehow Concourse put.inputs are like global variables: you don't know if and where they will be used in the job :-/ I will see how to clarify.

cogito/gchatsink.go
```diff
@@ -12,9 +13,10 @@ import (
 
 // GoogleChatSink is an implementation of [Sinker] for the Cogito resource.
 type GoogleChatSink struct {
-	Log     hclog.Logger
-	GitRef  string
-	Request PutRequest
+	Log      hclog.Logger
+	InputDir fs.FS
```
aliculPix4D, line 17: Totally unrelated to this PR, but just for my curiosity.

I forgot, but can you remind be why we were using `afero` in `fly_helper`. It was something about read/write and `fs.FS` supporting only read, but how you are using the `fstest` here, I am not sure anymore if `afero` was needed. 

What do we write with fly_helper that we (and by this I mean I :) ) decided for afero?

marco-m-pix4d: Maybe we write nothing :-D I do not remember :-D

aliculPix4D: I remember the plan for `fly_helper init` command, but for sure this was not the reason. I will have a look again.

pipelines/cogito-acceptance.yml
marco-m-pix4d: Each time I am confused, so I look at the table in the Cogito README and I _assume_ that it is correct. I created it based on Concourse documentation:

- failure | 🔴 - task exited non 0
- error | 🟠 - any other error besides failure or abort (pipeline configuration error, network error, timeout, ...)

it is true that the resource itself if it exits non 0 it is reported red/failure, but I _assume_ that the job becomes orange/error. I might be wrong.

Taskfile.yml
```diff
@@ -117,6 +119,43 @@ tasks:
...
+    desc: Run the Cogito acceptance tests. Needs a running Concourse.
+    cmds:
+      - task: test:acceptance:set-pipeline
+      - task: test:acceptance:chat-message-file
+      - task: test:acceptance:chat-message-append
+
+  test:acceptance:fly-login:
+    desc: Performs a fly login in the target to be used in the acceptance tests.
+    cmds:
+      - fly -t cogito login -c $(gopass show cogito/concourse_url) --open-browser
```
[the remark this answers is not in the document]
marco-m-pix4d: done

cogito/protocol_test.go
```diff
@@ -289,9 +294,12 @@ gchat_webhook: ***REDACTED***`
 		input := cogito.PutParams{
 			State: cogito.StateFailure,
 		}
-		want := `state:         failure
-context:       
-gchat_webhook: `
+		want := `state:               failure
+context:
```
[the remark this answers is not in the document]
marco-m-pix4d: done

Taskfile.yml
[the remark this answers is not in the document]
marco-m-pix4d: Done.

README.md
marco-m-pix4d: Updated.

cogito/put.go
marco-m-pix4d: @Esysc what is the "block"? Only lines 168--171 or more?

cogito/put.go
marco-m-pix4d: @odormond with a fresh mind I now think I understand your proposal. Seems nice indeed. Will try it, thanks.

marco-m-pix4d: @odormond the last part
```
if inputDirs.size() != 1 {
  return fmt.Errorf("blabla missing git repo")
}
```
was not enough. I had to change it to:
```
	if inputDirs.Size() == 0 {
		return fmt.Errorf(
			"put:inputs: missing directory for GitHub repo: have: %v, GitHub: %s/%s",
			collected, source.Owner, source.Repo)
	} else if inputDirs.Size() > 1 {
		return fmt.Errorf(
			"put:inputs: want only directory for GitHub repo: have: %v, GitHub: %s/%s",
			inputDirs, source.Owner, source.Repo)
	}
```

It might still be more readable than what I had.

marco-m-pix4d:
@Pix4D/integration review comments addressed in two separate commits (f3729e2 and 855c819) to keep them as clear as possible.

Ready for another review, thanks.

odormond (approved):

cogito/put.go
Esysc: Never mind, I see you addressed in https://github.com/Pix4D/cogito/pull/100/commits/855c81916ffa98c8c856b055347714e1c0a99fc5

Esysc (approved):

iAmoric (approved):

PR #101, opened 2022-09-01
Chat: understand user intent

marco-m-pix4d:
This fixes the limitation I mentioned this morning in the standup, that @sambarluc and myself discovered in a test pipeline with the previous release candidate.

Before this change, if on a specific put the user wanted to send a notification
message, for example:
```yaml
    - put: cogito
      inputs: [repo.git, app]
      params:
        state: success
        chat_message: "Perftracker registration of X"
```
He would be forced to actually make the pipeline spammy by adding state
`success` to `source.chat_notify_on_states`.

To solve this, we had two options:

1. Add `put.params.chat_notify_on_states`, but this would look ugly:
```yaml
    - put: cogito
      inputs: [repo.git, app]
      params:
        state: success
        chat_notify_on_states: [success]        <== well it is obvious, no?
        chat_message: "Perftracker registration of X"
```
2. Understand what the user wants: it is obvious that at least one of
chat_message or chat_message_file is set, then the user wants to send the chat
message in this specific put. Note the specific == NON in a job on_xxx handler.

In this PR we implement with option 2.

### DIFF
--- a/cogito/gchatsink.go
+++ b/cogito/gchatsink.go
@@ -36,7 +36,7 @@ func (sink GoogleChatSink) Send() error {
 	}
 
 	state := sink.Request.Params.State
-	if !shouldSendToChat(state, sink.Request.Source.NotifyOnStates) {
+	if !shouldSendToChat(state, sink.Request.Source.ChatNotifyOnStates) {
 		sink.Log.Debug("not sending to chat",
 			"reason", "state not in enabled states", "state", state)
 		return nil
--- a/cogito/gchatsink.go
+++ b/cogito/gchatsink.go
@@ -36,9 +36,9 @@ func (sink GoogleChatSink) Send() error {
 	}
 
 	state := sink.Request.Params.State
-	if !shouldSendToChat(state, sink.Request.Source.ChatNotifyOnStates) {
+	if !shouldSendToChat(sink.Request) {
 		sink.Log.Debug("not sending to chat",
-			"reason", "state not in enabled states", "state", state)
+			"reason", "state not in configured states", "state", state)
 		return nil
 	}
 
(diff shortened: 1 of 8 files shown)


cogito/gchatsink.go
```diff
@@ -36,9 +36,9 @@ func (sink GoogleChatSink) Send() error {
 	}
 
 	state := sink.Request.Params.State
-	if !shouldSendToChat(state, sink.Request.Source.NotifyOnStates) {
+	if !shouldSendToChat(sink.Request) {
 		sink.Log.Debug("not sending to chat",
-			"reason", "state not in enabled states", "state", state)
+			"reason", "state not in configured states", "state", state)
```
Esysc, line 41: I'm confused here. It seems that `shouldSendToChat` now verifies also the presence of a chat content, not only the state.

marco-m-pix4d: Yes, it is explained in the PR description and in the README changes in this PR.

sambarluc:
I might have messed up something in the configuration in here https://github.com/Pix4D/analytics-inspect-antdet/pull/227 but it doesn't seem to work as expected: https://builder.ci.pix4d.com/teams/developers/pipelines/analytics-inspect-antdet-PCI-2561-test-new-cogito/jobs/perftracker-registration/builds/10#L631162bf:68

marco-m-pix4d:
@sambarluc I left a (I hope) explanation in the test PR.

cogito/gchatsink.go
Esysc: Yes it is. Now you could have 2 reasons to skip the chat.

marco-m-pix4d: This is what I though at the beginning also, and did another implementation of `shouldSendToChat` that itself returned the reason (since it knows everything).

At the end, I decided it was not worth it, at least for the time being, because this reason "state not in configured states" is always true. It could say instead "state not in configured states and both chat_message and chat_message_file are empty", but to me it would just add confusion...

Esysc: I understand, so let keep it as is.

Esysc (approved):

aliculPix4D (approved):

PR #102, opened 2022-09-06
chat: append the build summary by default

marco-m-pix4d:
chat: append the build summary by default

Configuration:

- `source.chat_append_build_summary` (default: `true`)
- `put.params.chat_append_build_summary` (default: `source.chat_append_build_summary`)

Once this is merged, we can release :-)

A good way to get an overview is to look at the acceptance pipeline: pipelines/cogito-acceptance.yml

PCI-2663

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,14 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [v0.8.1] - UNRELEASED
 
+This release allows to use cogito for all chat notifications where previously one would have needed a mix of cogito and concourse-hangouts-resource.
+
 ### Added
 
-- Google Chat: allow selecting which states trigger a notification to the chat, via configuration `source.chat_notify_on_states`. Default (as before this tunable): [`abort`, `error`, `failure`].
+- Google Chat: allow selecting which states trigger a notification to the chat, via configuration `source.chat_notify_on_states`. Default (as before this tunable): `[abort, error, failure]`.
 - Google Chat: enrich cogito logging with human-readable information about the chat space.
-- Google Chat: allow to specify a different chat space/webhook during put (see `params.gchat_webhook`).
-- Google Chat: allow to specify a custom chat message, overriding the default build summary (see `params.chat_message`).
-- Google Chat: when passing a custom message or a custom message file, choose whether to append to it the default build summary or not (see `params.chat_message_append`).
-- Google Chat: allow to specify a file containing a custom chat message (see `params.chat_message_file`).
+- Google Chat: allow to specify a different chat space/webhook during put (see `put.params.gchat_webhook`).
+- Google Chat: allow to specify a custom chat message, (see `params.chat_message`).
+- Google Chat: allow to specify a file containing a custom chat message (see `put.params.chat_message_file`).
+- Google Chat: when passing a custom message or a custom message file, choose whether to append to it the default build summary or not (see `source.chat_append_summary`, `put.params.chat_append_summary`). Default: `true`.
 
 ## [v0.8.0] - 2022-08-11
 
--- a/cogito/protocol.go
+++ b/cogito/protocol.go
@@ -4,6 +4,7 @@
 package cogito
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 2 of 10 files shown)


marco-m-pix4d:
FYI @sambarluc @klaussnd

cogito/protocol.go
```diff
@@ -48,6 +49,39 @@ type PutRequest struct {
 	Env    Environment
 }
 
+func (req *PutRequest) UnmarshalJSON(data []byte) error {
+	type request PutRequest // Alias to avoid infinite loops.
```
marco-m-pix4d, line 53: See also
- https://www.orsolabs.com/post/go-json-default-values/
- https://eli.thegreenplace.net/2020/optional-json-fields-in-go/

marco-m-pix4d: To wrap your mind around this, start with Source.UnmarshalJSON() at line 126 below.

aliculPix4D (approved):

CHANGELOG.md
```diff
@@ -7,14 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Google Chat: enrich cogito logging with human-readable information about the chat space.
-- Google Chat: allow to specify a different chat space/webhook during put (see `params.gchat_webhook`).
-- Google Chat: allow to specify a custom chat message, overriding the default build summary (see `params.chat_message`).
-- Google Chat: when passing a custom message or a custom message file, choose whether to append to it the default build summary or not (see `params.chat_message_append`).
-- Google Chat: allow to specify a file containing a custom chat message (see `params.chat_message_file`).
+- Google Chat: allow to specify a different chat space/webhook during put (see `put.params.gchat_webhook`).
+- Google Chat: allow to specify a custom chat message, (see `params.chat_message`).
+- Google Chat: allow to specify a file containing a custom chat message (see `put.params.chat_message_file`).
+- Google Chat: when passing a custom message or a custom message file, choose whether to append to it the default build summary or not (see `source.chat_append_summary`, `put.params.chat_append_summary`).
```
iAmoric, line 19: You don't want to specify the default value in the changelog? (even if it's in the readme) Like you did for `source.chat_notify_on_states`?

iAmoric (approved):

cogito/protocol.go
```diff
@@ -79,18 +114,33 @@ func (src Source) String() string {
...
-	return s
+	// Since we also want to enforce the parser to fail if it encounters unknown fields,
+	// we cannot use the customary json.Unmarshal(data, &aux) but we have to go through
+	// a json decoder :-/
+	dec := json.NewDecoder(bytes.NewReader(data))
+	dec.DisallowUnknownFields()
+	if err := dec.Decode(&aux); err != nil {
+		return err
+	}
+	*src = Source(aux)
```
aliculPix4D, line 141: @marco-m-pix4d I have a stupid question regarding this line. 

Source is `type Source struct` and aux is of the same type, so doing `Source(aux)` is some kind of type checking?
Is this similar to `int(6)` or `int(6.3)`?

odormond (approved):

cogito/protocol.go
marco-m-pix4d: It is not obvious for sure! Your explanation is correct.

You see? One day you will even start appreciating Go :trollface:

CHANGELOG.md
marco-m-pix4d: Ah, good point.

cogito/protocol.go
marco-m-pix4d: In Go it is called a "type conversion", see for a simple example https://go.dev/tour/basics/13.
This is different from "type assert" (https://go.dev/tour/methods/15).

Also @odormond explanation above should help.

With 
```go
*src = aux
```
we would get
```
cannot use aux (variable of type source) as type Source in assignment
```

cogito/protocol.go
aliculPix4D: Just to clarify additionally for me one more thing:
It does not find it because the function signature are different?
`func (req *PutRequest) UnmarshalJSON(data []byte) error`
vs
`func (request PutRequest) UnmarshalJSON(data []byte) error`
?

CHANGELOG.md
marco-m-pix4d: updated

cogito/protocol.go
odormond: Almost. The other one would not be `func (request PutRequest) UnmarshalJSON(data []byte) error` but `func (req *request) UnmarshalJSON(data []byte) error`

PR #104, opened 2022-09-08
peek at the log level and simplify object construction

marco-m-pix4d:
This is one of the cleanups I mentioned at the standup

Better reviewed commit-per-commit because a lot of file overlaps.

TODO

- [x] remove silient/off backward compatibility

### DIFF
--- a/cmd/cogito/main.go
+++ b/cmd/cogito/main.go
@@ -71,7 +71,7 @@ func mainErr(in io.Reader, out io.Writer, logOut io.Writer, args []string) error
 
 // peekLogLevel decodes 'input' as JSON and looks for key source.log_level. If 'input'
 // is not JSON, peekLogLevel will return an error. If 'input' is JSON but does not
-// contain key source.log_level, peekLogLevel returns the empty string.
+// contain key source.log_level, peekLogLevel returns "info" as default value.
 //
 // Rationale: depending on the Concourse step we are invoked for, the JSON object we get
 // from stdin is different, but it always contains a struct with name "source", thus we
@@ -83,7 +83,7 @@ func peekLogLevel(input []byte) (string, error) {
 		} `json:"source"`
 	}
 	var peek Peek
-	peek.Source.LogLevel = "info"
+	peek.Source.LogLevel = "info" // default value
 	if err := json.Unmarshal(input, &peek); err != nil {
 		return "", fmt.Errorf("peeking into JSON for log_level: %s", err)
 	}
--- a/cmd/cogito/main.go
+++ b/cmd/cogito/main.go
@@ -3,6 +3,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -10,6 +11,7 @@ import (
 
 	"github.com/Pix4D/cogito/cogito"
 	"github.com/Pix4D/cogito/github"
+	"github.com/Pix4D/cogito/sets"
 	"github.com/hashicorp/go-hclog"
 )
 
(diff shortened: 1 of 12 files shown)


cmd/cogito/main.go
```diff
@@ -18,15 +20,30 @@ func main() {
...
+	validCmds := sets.From("check", "in", "out")
+	if !validCmds.Contains(cmd) {
+		return fmt.Errorf("invoked as '%s'; want: one of %v", cmd, validCmds)
+	}
+
+	input, err := io.ReadAll(in)
+	if err != nil {
+		return fmt.Errorf("reading stdin: %s", err)
+	}
+	logLevel, err := peekLogLevel(input)
```
marco-m-pix4d, line 40: The main idea is here: peek at the log_level from main.
To be able to then parse the input again, I also had to read the stdin stream once into a buffer and pass it around.

cogito/protocol.go
```diff
@@ -143,41 +209,6 @@ func (src *Source) UnmarshalJSON(data []byte) error {
...
 }
 
-// ValidateLog validates and applies defaults for the logging configuration of Source.
-//
-// This chicken-and-egg problem is due to the fact that logging configuration is passed
-// too late, at the same time as all the other resource Source configuration, so to give
-// as much debugging information as possible we need to get the log level as soon as
-// possible, also if the Source has other errors. This cannot be simplified, we are
-// working within the limits of the Concourse resource protocol.
-func (src *Source) ValidateLog() error {
```
marco-m-pix4d, line 153: this is now directly in ValidateSource

cogito/put.go
```diff
@@ -94,30 +94,18 @@ func NewPutter(ghAPI string, log hclog.Logger) *ProdPutter {
...
-	if err := putter.Request.Source.Validate(); err != nil {
-		return fmt.Errorf("put: %s", err)
+	request, err := NewPutRequest(input)
+	if err != nil {
+		return err
 	}
-
-	putter.log.Debug("after validation and defaults",
+	putter.Request = request
+	putter.log.SetLevel(hclog.LevelFromString(request.Source.LogLevel))
```
odormond, line 107: Why do you re-set the log level? It's already set to the value of `request.Source.LogLevel` on line 46 of `cogito/main.go`, isn't it?

cogito/protocol.go
```diff
@@ -49,6 +94,27 @@ type PutRequest struct {
 	Env    Environment
 }
 
+// NewPutRequest returns a [PutRequest] ready to be used.
+func NewPutRequest(input []byte) (PutRequest, error) {
```
Esysc, line 98: So we are forced to use multiple functions because the struct differs.

cogito/put.go
marco-m-pix4d: I should have added a comment indeed. I am also not 100% sure of what to do.

In main we peek at the log level but in a opaque way, to allow to go on and give more logging before failing.
If at that moment it is set to "banana", such value will be accepted and passed to the logging system, which will swallow it, recognize it as invalid and decide to set the level as INFO.

When we arrive here at line 107, the level has been validated (in Source.Validate, in this PR), so it seems it is indeed useless to re-assign. The reason why we are doing it is because of the mapping (see Source.Validate):
```go
	"silent": "off", // different
```
When we were using the ofcourse package, we published `silent` as a valid log level.
If a pipeline is using `silent`, without the re-assignment of line 107, silent would have been recognized as invalid and then INFO in main.
With this 2nd assignment, the level would be set to `off`.

So this is a good question, because it allows me to ask: is it worthwhile to go to the current confusion (well, before this PR it was even way worse) just for this theoretical backward compatibility?

I am now tempted to say no. If a pipeline really exists out there with `silent`, it will simply start to log at INFO. If the owner notices, he/she simply has to replace `silent` with `off` in the pipeline.

Another possibility that I even considered and that might be even better is to reduce the levels to simply INFO and DEBUG, for two reasons:

1. the code only uses INFO or DEBUG and will keep doing this; the other levels do not make sense.
2. I cannot see a valid case to set the logging to `off`...

Opinions?

cogito/protocol.go
marco-m-pix4d: Yes, each time the JSON input is partially different. "Source" is always there but not the rest.
See https://concourse-ci.org/implementing-resource-types.html

cogito/put.go
Esysc: So, as you explained in another PR, you assign a log level before to parse the source, then you tune with what is defined in `source`.

odormond: > Opinions?

I would just drop the `silence`/`off` support. It's not like they'll see the logs in their jobs. They'll have to go look at the resource steps to see them. If they do so, then what's the point of having turned logging off?

marco-m-pix4d: This looks like a reasonable compromise indeed.

cogito/protocol.go
aliculPix4D: For my understanding why the difference:
`func NewGetRequest(input []byte) (*GetRequest, error) {`
vs
`func NewPutRequest(input []byte) (PutRequest, error) {`
the first one returns the pointer, but the second full structure?

aliculPix4D (approved):
LGTM; but as suggested for logging, I would support only INFO and DEBUG, since as you said this are the only one used in the code.

cogito/protocol.go
aliculPix4D: Thanks for the comments and links. I am going through them now.

> So in this case, `func NewGetRequest() *GetRequest` returns a pointer to the memory allocated (on the heap) (in C this would crash instead) from the function, and `func NewPutRequest() PutRequest` returns a copy of the memory allocated (on the stack) from the function.

But I still don't get why the difference in this code? Why this "inconsistency"?

marco-m-pix4d: > But I still don't get why the difference in this code?

I misunderstood the question. Now I see. There is no reason indeed, so I am adding a commit to remove the difference, thanks for asking.

cogito/put.go
marco-m-pix4d: Dropped the silence -> off support in 77d369f

Esysc (approved):

odormond (approved):

PR #109, opened 2022-11-14
Pci 2665 independent chat send add sinks to source

Esysc:
Part of: [PCI-2665](https://pix4dbug.atlassian.net/browse/PCI-2665)

The resource Source configuration now parses the `sinks` parameter.

Example:

Obviously the example below is not realistic, we don't need to specify the default.
```yaml
resources:
- name: gh-status
  type: cogito
  check_every: never
  source:
    owner: ((github-owner))
    repo: ((your-repo-name))
    access_token: ((github-PAT))
    sinks: ["github", "gchat"]
```

A more realistic example:
```yaml
resources:
- name: gchat
  type: cogito
  check_every: never
  source:
    sinks: ["gchat"],
    gchat_webhook: https://my-chat-webhook
```
Note that when specifying the sink and only chat feature, the GitHub parameter are not mandatory anymore, but the webhook will be.

Best to review commit by commit.

### Todos

- [ ] Update documentation (separate PR)
- [ ] make a new relase

I didn't touch the documentation yet, I prefer to wait for your feedback.

### DIFF
--- a/cogito/check_test.go
+++ b/cogito/check_test.go
@@ -99,10 +99,10 @@ func TestCheckFailure(t *testing.T) {
 		{
 			name: "validation failure: wrong sink key",
 			source: cogito.Source{
-				Sinks: []string{"slack", "gchat"},
+				Sinks: []string{"ghost", "gchat"},
 			},
 			writer:  io.Discard,
-			wantErr: "check: source: invalid sink: [slack]. Supported sinks: [gchat github]",
+			wantErr: "check: source: invalid sink(s): [ghost]. Supported sinks: [gchat github]",
 		},
 		{
 			name:    "write error",
--- a/cogito/check_test.go
+++ b/cogito/check_test.go
@@ -53,6 +53,16 @@ func TestCheckSuccess(t *testing.T) {
 			},
 			wantOut: []cogito.Version{{Ref: "dummy"}},
 		},
+		{
+			name: "first request (Concourse omits the version field) gchat only",
+			request: cogito.CheckRequest{
+				Source: cogito.Source{
+					Sinks:        []string{"gchat"},
+					GChatWebHook: "https://dummy-webhook",
+				},
+			},
+			wantOut: []cogito.Version{{Ref: "dummy"}},
+		},
 	}
 
 	for _, tc := range testCases {
@@ -88,10 +98,18 @@ func TestCheckFailure(t *testing.T) {
 
 	testCases := []testCase{
 		{
-			name:    "validation failure: missing keys",
+			name:    "validation failure: missing repo keys",
 			source:  cogito.Source{},
 			writer:  io.Discard,
 			wantErr: "check: source: missing keys: owner, repo, access_token",
+		},
+		{
+			name: "validation failure: missing gchat keys",
+			source: cogito.Source{
+				Sinks: []string{"gchat"},
+			},
+			writer:  io.Discard,
+			wantErr: "check: source: missing keys: gchat_webhook",
 		},
 		{
 			name:    "write error",
--- a/cogito/check_test.go
+++ b/cogito/check_test.go
@@ -31,24 +31,18 @@ func TestCheckSuccess(t *testing.T) {
 		assert.DeepEqual(t, have, tc.wantOut)
 	}
 
-	baseSource := cogito.Source{
-		Owner:       "the-owner",
-		Repo:        "the-repo",
-		AccessToken: "the-token",
-	}
-
 	testCases := []testCase{
 		{
 			name: "first request (Concourse omits the version field)",
 			request: cogito.CheckRequest{
-				Source: baseSource,
+				Source: baseGithubSource,
 			},
 			wantOut: []cogito.Version{{Ref: "dummy"}},
 		},
 		{
 			name: "subsequent requests (Concourse adds the version field)",
 			request: cogito.CheckRequest{
-				Source:  baseSource,
+				Source:  baseGithubSource,
 				Version: cogito.Version{Ref: "dummy"},
 			},
 			wantOut: []cogito.Version{{Ref: "dummy"}},
@@ -56,10 +50,7 @@ func TestCheckSuccess(t *testing.T) {
 		{
 			name: "first request (Concourse omits the version field) gchat only",
 			request: cogito.CheckRequest{
-				Source: cogito.Source{
-					Sinks:        []string{"gchat"},
-					GChatWebHook: "https://dummy-webhook",
-				},
+				Source: baseGchatSource,
 			},
 			wantOut: []cogito.Version{{Ref: "dummy"}},
 		},
@@ -90,12 +81,6 @@ func TestCheckFailure(t *testing.T) {
 		assert.Error(t, err, tc.wantErr)
 	}
 
-	baseSource := cogito.Source{
-		Owner:       "the-owner",
-		Repo:        "the-repo",
-		AccessToken: "the-token",
-	}
-
 	testCases := []testCase{
 		{
 			name:    "validation failure: missing repo keys",
@@ -113,7 +98,7 @@ func TestCheckFailure(t *testing.T) {
 		},
 		{
 			name:    "write error",
-			source:  baseSource,
+			source:  baseGithubSource,
 			writer:  &testhelp.FailingWriter{},
 			wantErr: "check: preparing output: test write error",
 		},
--- a/cogito/check_test.go
+++ b/cogito/check_test.go
@@ -97,6 +97,14 @@ func TestCheckFailure(t *testing.T) {
 			wantErr: "check: source: missing keys: gchat_webhook",
 		},
 		{
+			name: "validation failure: wrong sink key",
+			source: cogito.Source{
+				Sinks: []string{"slack", "gchat"},
+			},
+			writer:  io.Discard,
+			wantErr: "check: source: invalid sink: [slack]. Supported sinks: [gchat github]",
+		},
+		{
 			name:    "write error",
 			source:  baseGithubSource,
 			writer:  &testhelp.FailingWriter{},
--- a/cogito/get_test.go
+++ b/cogito/get_test.go
@@ -116,6 +116,14 @@ func TestGetFailure(t *testing.T) {
 			writer:  &testhelp.FailingWriter{},
 			wantErr: "get: source: missing keys: gchat_webhook",
 		},
+		{
+			name: "validation failure: wrong sink key",
+			source: cogito.Source{
+				Sinks: []string{"slack", "gchat"},
+			},
+			writer:  io.Discard,
+			wantErr: "get: source: invalid sink: [slack]. Supported sinks: [gchat github]",
+		},
 	}
 
 	for _, tc := range testCases {
--- a/cogito/get_test.go
+++ b/cogito/get_test.go
@@ -119,10 +119,10 @@ func TestGetFailure(t *testing.T) {
 		{
 			name: "validation failure: wrong sink key",
 			source: cogito.Source{
-				Sinks: []string{"slack", "gchat"},
+				Sinks: []string{"ghost", "gchat"},
 			},
 			writer:  io.Discard,
-			wantErr: "get: source: invalid sink: [slack]. Supported sinks: [gchat github]",
+			wantErr: "get: source: invalid sink(s): [ghost]. Supported sinks: [gchat github]",
 		},
 	}
 
--- a/cogito/get_test.go
+++ b/cogito/get_test.go
@@ -31,22 +31,11 @@ func TestGetSuccess(t *testing.T) {
 		assert.DeepEqual(t, have, tc.wantOut)
 	}
 
-	baseGitSource := cogito.Source{
-		Owner:       "the-owner",
-		Repo:        "the-repo",
-		AccessToken: "the-token",
-	}
-
-	baseGchatSource := cogito.Source{
-		Sinks:        []string{"gchat"},
-		GChatWebHook: "https://the-webhook",
-	}
-
 	testCases := []testCase{
 		{
 			name: "returns requested version",
 			request: cogito.GetRequest{
-				Source:  baseGitSource,
+				Source:  baseGithubSource,
 				Version: cogito.Version{Ref: "banana"},
 			},
 			wantOut: cogito.Output{Version: cogito.Version{Ref: "banana"}},
@@ -92,12 +81,6 @@ func TestGetFailure(t *testing.T) {
 		assert.Error(t, err, tc.wantErr)
 	}
 
-	baseSource := cogito.Source{
-		Owner:       "the-owner",
-		Repo:        "the-repo",
-		AccessToken: "the-token",
-	}
-
 	testCases := []testCase{
 		{
 			name:    "user validation failure: missing keys",
@@ -107,13 +90,13 @@ func TestGetFailure(t *testing.T) {
 		},
 		{
 			name:    "concourse validation failure: empty version field",
-			source:  baseSource,
+			source:  baseGithubSource,
 			writer:  io.Discard,
 			wantErr: "get: empty 'version' field",
 		},
 		{
 			name:    "concourse validation failure: missing output directory",
-			source:  baseSource,
+			source:  baseGithubSource,
 			version: cogito.Version{Ref: "dummy"},
 			writer:  io.Discard,
 			wantErr: "get: arguments: missing output directory",
@@ -121,7 +104,7 @@ func TestGetFailure(t *testing.T) {
 		{
 			name:    "system write error",
 			args:    []string{"dummy-dir"},
-			source:  baseSource,
+			source:  baseGithubSource,
 			version: cogito.Version{Ref: "dummy"},
 			writer:  &testhelp.FailingWriter{},
 			wantErr: "get: preparing output: test write error",
--- a/cogito/get_test.go
+++ b/cogito/get_test.go
@@ -31,17 +31,30 @@ func TestGetSuccess(t *testing.T) {
 		assert.DeepEqual(t, have, tc.wantOut)
 	}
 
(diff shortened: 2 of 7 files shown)


cogito/protocol.go
```diff
@@ -212,18 +216,38 @@ func (src *Source) UnmarshalJSON(data []byte) error {
 // Validate verifies the Source configuration and applies defaults.
 func (src *Source) Validate() error {
 	//
-	// Validate mandatory fields.
+	// Evaluate mandatory fields.
```
Esysc, line 219: As stated before, `Validate()` is dynamic: some fields will become mandatory/optional based on the requested sinks.

cogito/putter.go
```diff
@@ -152,7 +168,12 @@ func (putter *ProdPutter) Sinks() ([]Sinker, error) {
 			Request:  putter.Request,
 		},
 	}
-	sinksParams := putter.Request.Params.Sinks
+	// Sinks configured in Put params get priority.
```
Esysc, line 171: Resource Source configuration can be overridden in the `put` step parameters.

cogito/put_test.go
```diff
@@ -186,11 +186,40 @@ func TestPutterLoadConfigurationInvalidParamsFailure(t *testing.T) {
...
+	err := putter.LoadConfiguration(in, nil)
+
+	assert.Error(t, err, wantErr)
+}
+
+func TestPutterLoadConfigurationUnknownSink(t *testing.T) {
+	in := []byte(`
+{
+  "source": {"sinks": ["pizza"]},
+  "params": {}
```
odormond, line 207: You should do a similar test but with the sinks set in the params.

odormond: Also, I haven't found a test checking that the `sinks` configured in the `source` can be overwritten by the ones in `params`.

cogito/check_test.go
```diff
@@ -53,6 +53,16 @@ func TestCheckSuccess(t *testing.T) {
 			},
 			wantOut: []cogito.Version{{Ref: "dummy"}},
 		},
+		{
+			name: "first request (Concourse omits the version field) gchat only",
+			request: cogito.CheckRequest{
+				Source: cogito.Source{
+					Sinks:        []string{"gchat"},
```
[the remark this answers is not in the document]
odormond: You should also test what happens with an unknown sink being configured to make sure the error message is understandable.

cogito/get_test.go
```diff
@@ -31,17 +31,30 @@ func TestGetSuccess(t *testing.T) {
 		assert.DeepEqual(t, have, tc.wantOut)
 	}
 
-	baseSource := cogito.Source{
+	baseGitSource := cogito.Source{
```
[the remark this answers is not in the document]
Esysc: Thanks, it's done. I also put this global test variables in a separate file.

cogito/protocol.go
```diff
@@ -212,18 +216,38 @@ func (src *Source) UnmarshalJSON(data []byte) error {
...
+		if !sinksSet.Contains("github") {
+			isGitMandatory = false
+		}
+		if sinksSet.Contains("gchat") {
+			isGchatOptional = false
+		}
+		if !(sinksSet.Contains("gchat") || sinksSet.Contains("github")) {
+			return fmt.Errorf("source: invalid sink: %s. Supported sinks: 'gchat', 'github'", sinksSet)
+		}
+	}
```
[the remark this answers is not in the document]
Esysc: Thanks for the suggestion, I've partially applied what you proposed. Basically I need to check every Source parameter that is marked as mandatory, one of them may be correctly filled while another left empty, so we return a clear message to the user about what is missing.

cogito/protocol.go
```diff
@@ -212,18 +217,47 @@ func (src *Source) UnmarshalJSON(data []byte) error {
...
+	if sinksSet.Size() > 0 {
+		sinksNotValid := sinksSet.Difference(defaultSinksSet)
+		if sinksNotValid.Size() > 0 {
+			return fmt.Errorf("source: invalid sink: %s. Supported sinks: %s", sinksNotValid, defaultSinks)
+		}
 	}
-	if src.Repo == "" {
-		mandatory = append(mandatory, "repo")
+
+	v := reflect.ValueOf(src).Elem()
```
aliculPix4D, line 240: I think @marco-m-pix4d will not like this :)

Esysc: May be, I tried to address the suggestion from @odormond , may be I could obtain the same differently.

cogito/protocol.go
```diff
@@ -212,18 +217,47 @@ func (src *Source) UnmarshalJSON(data []byte) error {
...
+		"Repo":        "repo",
+		"AccessToken": "access_token",
+	}
+
+	defaultSinks := []string{"gchat", "github"}
+	defaultSinksSet := sets.From(defaultSinks...)
+	sinksSet := sets.From(src.Sinks...)
+
+	if sinksSet.Size() > 0 {
+		sinksNotValid := sinksSet.Difference(defaultSinksSet)
```
aliculPix4D, line 234: in my understanding this code segment checks if the configured sinks are allowed.

So for me this is not defaultSinks its allowedSinks (or supported).

Additionally, at how many places we will need a list of the defaultSinks? 

This should be a global variable (`[]string{"gchat", "github"}`) which we change once and reuse.

cogito/protocol.go
```diff
@@ -212,18 +216,43 @@ func (src *Source) UnmarshalJSON(data []byte) error {
...
+		// First validate sinks are known and supported.
+		sinksNotValid := sinksSet.Difference(defaultSinksSet)
+		if sinksNotValid.Size() > 0 {
+			return fmt.Errorf("source: invalid sink: %s. Supported sinks: %s", sinksNotValid, defaultSinks)
+		}
 	}
-	if src.Repo == "" {
-		mandatory = append(mandatory, "repo")
+
+	if sinksSet.Size() == 0 || sinksSet.Contains("github") {
```
aliculPix4D, line 235: Ufff I was just starting to review the previous code with `reflect`. This is much clearer to review :)

cogito/protocol.go
Esysc: Something similar as I did in `vars_test,go`? I thought about it, so we will have just a centralized point to declare the supported sinks.

cogito/protocol.go
```diff
@@ -212,18 +216,43 @@ func (src *Source) UnmarshalJSON(data []byte) error {
...
+			mandatory = append(mandatory, "repo")
+		}
+		if src.AccessToken == "" {
+			mandatory = append(mandatory, "access_token")
+		}
 	}
-	if src.AccessToken == "" {
-		mandatory = append(mandatory, "access_token")
+
+	if sinksSet.Size() > 0 && sinksSet.Contains("gchat") {
```
aliculPix4D, line 248: This is a breaking change compared to current behavior, no? If you want to use `github` and `gchat`, now you are required to set the sinks explicitly in source...

what happens if you don't set any sinks and you want use both `github` and `gchat`?

Or am I missing something here?

aliculPix4D: Or I was wrong, if we don't set any sinks webhook is just optional param.

Esysc: It's 100% backward compatible, the default is applied, that means `github, gchat` are available. Note that this validation runs around the source configuration and not the put parameters. 

If you don't specify any sink, `repo,owner,access_token` are mandatory because Cogito default to its original role. In the other hand if you specify `gchat` you are requested to set the webhook.

cogito/putter.go
```diff
@@ -51,13 +51,15 @@ func (putter *ProdPutter) LoadConfiguration(input []byte, args []string) error {
...
 	// args[0] contains the path to a directory containing all the "put inputs".
-	if len(args) == 0 {
+	if len(args) == 0 && len(sinks) == 0 {
 		return fmt.Errorf("put: arguments: missing input directory")
 	}
-	putter.InputDir = args[0]
-	putter.log.Debug("", "input-directory", putter.InputDir)
-
+	if len(args) > 0 {
+		putter.InputDir = args[0]
```
aliculPix4D, line 60: For my understanding: why do we expect that git repo is the first arg `args[0]`?

cogito/check_test.go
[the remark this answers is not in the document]
Esysc: This is covered now

Esysc:
Integration tests are failing because of the test repo commit that has reached the max statuses, I'm fixing it.

aliculPix4D:
@Esysc very minor, but you have a typo in the PR title. indipendent ---> independent

cogito/get_test.go
```diff
@@ -94,25 +90,40 @@ func TestGetFailure(t *testing.T) {
...
+			writer:  &testhelp.FailingWriter{},
+			wantErr: "get: source: missing keys: gchat_webhook",
+		},
+		{
+			name: "validation failure: wrong sink key",
+			source: cogito.Source{
+				Sinks: []string{"slack", "gchat"},
+			},
+			writer:  io.Discard,
+			wantErr: "get: source: invalid sink: [slack]. Supported sinks: [gchat github]",
```
aliculPix4D, line 125: same question as above in check tests. Any other name than slack?

cogito/protocol.go
```diff
@@ -212,18 +216,43 @@ func (src *Source) UnmarshalJSON(data []byte) error {
...
+
+	defaultSinks := []string{"gchat", "github"}
+	defaultSinksSet := sets.From(defaultSinks...)
+	sinksSet := sets.From(src.Sinks...)
+
+	if sinksSet.Size() > 0 {
+		// First validate sinks are known and supported.
+		sinksNotValid := sinksSet.Difference(defaultSinksSet)
+		if sinksNotValid.Size() > 0 {
+			return fmt.Errorf("source: invalid sink: %s. Supported sinks: %s", sinksNotValid, defaultSinks)
```
aliculPix4D, line 231: Plural: `invalid sinks:`. It can be more than one in my understanding.

cogito/protocol.go
aliculPix4D: Also, I would put this comment `// No sinks implies backward compatibility mode where github is mandatory and gchat optional.` above if statement not below.

cogito/protocol.go
aliculPix4D: I understand your point. Its still backward compatible. This is my understanding:

- so if no sinks are configured we go to default (current behavior) and `gchat_webhook` is optional. If it is set you will post to Gchat if it isn't set it will not post to Gchat.

- but than if Gchat is configured the `gchat_webhook` is mandatory.

So in both cases we can post to Gchat.

cogito/protocol_test.go
```diff
@@ -71,10 +65,23 @@ func TestSourceValidationFailure(t *testing.T) {
...
+		{
+			name:    "missing mandatory gchat source keys",
+			source:  cogito.Source{Sinks: []string{"gchat"}},
+			wantErr: "source: missing keys: gchat_webhook",
+		},
+		{
+			name: "invalid sink source keys",
+			source: cogito.Source{
+				Sinks:        []string{"gchat", "slacks"},
+				GChatWebHook: "sensitive-gchat-webhook",
```
aliculPix4D, line 81: why do you need this here and not before?

cogito/put_test.go
```diff
@@ -218,6 +257,26 @@ func TestPutterProcessInputDirSuccess(t *testing.T) {
...
+			sink:     "gchat",
+		},
+		{
+			name:     "only msg dir and message file, but gchat is set",
+			inputDir: "testdata/repo-and-msgdir/msgdir",
+			sink:     "gchat",
+			params:   cogito.PutParams{ChatMessageFile: "msgdir/msg.txt"},
+		},
+		{
+			name:   "sinks are correctly overrides by put params",
```
aliculPix4D, line 276: typo: overrides ---> overridden

cogito/protocol.go
Esysc: It could be, just for now we support only two, so it's rare to have more than one invalid.... I'll follow the suggestion.

cogito/putter.go
aliculPix4D: Ok its written above: `// args[0] contains the path to a directory containing all the "put inputs".`

but than this should be `putter.InputDirs` in plural no?

cogito/putter.go
```diff
@@ -51,13 +51,28 @@ func (putter *ProdPutter) LoadConfiguration(input []byte, args []string) error {
 		"environment", putter.Request.Env,
 		"args", args)
 
+	// If sink github is not requested, input dir is not mandatory.
+	// Validate the sinks first.
+	_, err = putter.Sinks()
+	if err != nil {
+		return fmt.Errorf("put: arguments: %w", err)
```
aliculPix4D, line 58: this does not tell a lot. `put: load config: failed to find sinks?`

cogito/check_test.go
```diff
@@ -80,22 +81,32 @@ func TestCheckFailure(t *testing.T) {
...
+			source: cogito.Source{
+				Sinks: []string{"gchat"},
+			},
+			writer:  io.Discard,
+			wantErr: "check: source: missing keys: gchat_webhook",
+		},
+		{
+			name: "validation failure: wrong sink key",
+			source: cogito.Source{
+				Sinks: []string{"slack", "gchat"},
```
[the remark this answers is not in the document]
Esysc: You'd change in any case, because the list of supported sinks should change when you add one.

cogito/get_test.go
Esysc: I changed, but be aware that you'll want to change the error message at every supported sinks change.

cogito/putter.go
```diff
@@ -83,14 +98,37 @@ func (putter *ProdPutter) ProcessInputDir() error {
...
+	// If putter.InputDir is not set, it is likely only Cogito chat feature is requested.
+	// Sinks can be set as Source, let check if they are configured and 'github' keyword
+	// is present. Also we return an error if 'sinks' is not configured in this case.
+	if inputDirs.Size() == 0 && (sinks.Size() == 0 || sinks.Contains("github")) {
+		return fmt.Errorf(
+			"put:inputs: missing directory for GitHub repo: have: %v, GitHub: %s/%s",
+			inputDirs, source.Owner, source.Repo)
+	}
+	// No errors, inputDirs is not mandatory at this point.
+	if inputDirs.Size() == 0 && (sinks.Size() > 0 && !sinks.Contains("github")) {
```
aliculPix4D, line 129: why do you need this check twice?

cogito/putter.go
```diff
@@ -83,14 +98,37 @@ func (putter *ProdPutter) ProcessInputDir() error {
 
 	params := putter.Request.Params
 	source := putter.Request.Source
+
+	sinks := sets.From(source.Sinks...)
+	if len(params.Sinks) > 0 {
```
aliculPix4D, line 103: How can it be that `sinks.Size() == 0`? Either `params.Sinks` or `source.Sinks` will return defaults no?

cogito/check_test.go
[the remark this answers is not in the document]
aliculPix4D: > because the list of supported sinks should change when you add one.

I think I wrote a comment in first PR about that. Global variable, set at single place. Overridable in test if your really wish.

cogito/putter.go
Esysc: The error is reported by `Sinks()` and as the name indicates, it's all about sinks.

cogito/putter.go
Esysc: Not exact, this is the pipeline configuration, the default is returned in case sinks are not configured. Here  by default we set sinks from `source` setting, but if specified in `put` parameter, then we override.

cogito/putter.go
Esysc: I will simplify grouping the common part that is `inputDirs.Size() == 0`.
 The first check, analyzes the `put` `inputs` and if there are no inputs you are in 2 possible scenarios:
1. if sinks are not specified or they are specified and `github` id found, this is an error, inputs are missing
2. if sinks are specified and `github` is not found, then it's ok we can return.

cogito/protocol_test.go
```diff
@@ -71,10 +65,23 @@ func TestSourceValidationFailure(t *testing.T) {
...
 			source:  cogito.Source{},
 			wantErr: "source: missing keys: owner, repo, access_token",
 		},
+		{
+			name:    "missing mandatory gchat source keys",
+			source:  cogito.Source{Sinks: []string{"gchat"}},
+			wantErr: "source: missing keys: gchat_webhook",
+		},
+		{
+			name: "invalid sink source keys",
```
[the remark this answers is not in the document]
Esysc: Added, also a mixed one.

cogito/protocol_test.go
aliculPix4D: Yes, just a few lines above, you don't mark anything as sensitive. In example L74 still uses Gchat but you don't configure `GChatWebHook`. Why is it needed in this test?

Esysc: Ah I understand now.
If you want to use the `gchat` feature of Cogito, `GChatWebHook`  becomes a mandatory parameter. In L74, we test just  that failure: `"source: missing keys: gchat_webhook"`.
Because Source validation comes before anything else, I need to specify `GChatWebHook` in the other places to test other error messages.  
About `sensitive-gchat-webhook` I'll change if it confuses, you can put what you want actually.

marco-m-pix4d:
@Esysc be careful that I forced-push your branches; you will have to do the following:
- commit any changes
- git fetch
- git rebase origin/pci-2665-indipendent-chat-send-add-sinks-to-source pci-2665-indipendent-chat-send-add-sinks-to-source

Esysc:
> 

All good, thanks @marco-m-pix4d

Esysc:
Closed, replaced by https://github.com/Pix4D/cogito/pull/113

PR #113, opened 2022-11-22
Cogito independent chat

Esysc:
Part of: [PCI-2665](https://pix4dbug.atlassian.net/browse/PCI-2665)

### Details
A new optional field is added and its name is `sinks`. It takes a slice of strings as values, corresponding to all Cogito supported sinks. 
This parameter can be set as `source.sinks` and/or `put.params.sinks` level. The latter if specified overrides the source configuration.  
In order to be backward compatible, if nothing is specified it behaviors as before defaulting to all supported sinks. 

Best review commit by commit.

### Tests criteria

image

The acceptance test, aka e2e tests, are increased by new CI jobs around the only chat feature:
* send message to chat using a standard Cogito resource, but specify the sink `gchat` in `put.params`.
* send message to chat using a custom only chat Cogito resource.
* send message to chat using a custom only chat Cogito resource plus setting a message directory and a message file.

image

The test coverage has slightly decreased (from 99.5% to 99.0%), I couldn't cover a special case using unit tests. Happily, it's covered by the acceptance tests.
Actually, the code below:

image

is executed clearly in the following CI job (see the red arrow):

image

### Todos
- [ ] Update remaining pipelines: replace Hangouts resource

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Cogito (**CO**ncourse **GIT** status res**O**urce) is a [Concourse resource] to update the GitHub commit status during a build. The name is a humble homage to [René Descartes].
 
-Optionally, it also sends a message to a chat system (currently supported: Google Chat). This allows to reduce the verbosity of a Concourse pipeline and especially to reduce the number of resource containers in a Concourse deployment, thus reducing load.
+It can also send a message to a chat system (currently supported: Google Chat). This allows to reduce the verbosity of a Concourse pipeline and especially to reduce the number of resource containers in a Concourse deployment, thus reducing load. Chat and GitHub commit status update can be used independently (see [examples](#examples) below).
 
 Written in Go, it has the following characteristics:
 
@@ -10,7 +10,7 @@ Written in Go, it has the following characteristics:
 - Extensive test suite.
 - Autodiscovery of configuration parameters.
 - No assumptions on the git repository (for example, doesn't assume that the default branch is `main` or that branch `main` even exists).
-- Supports Concourse 7.4 [instanced pipelines](https://concourse-ci.org/instanced-pipelines.html).
+- Supports Concourse [instanced pipelines](https://concourse-ci.org/instanced-pipelines.html).
 - Helpful error messages when something goes wrong with the GitHub API.
 - Configurable logging for the three steps (check, in, out) to help troubleshooting.
 
@@ -124,6 +124,20 @@ The absolute minimum is adding key `gchat_webhook` to the `source` configuration
     gchat_webhook: ((gchat_webhook))
 ```
 
+## Chat notification only
+
+The absolute minimum is setting keys `gchat_webhook` and `sinks` in the `source` configuration:
+
+```yaml
+- name: chat-notify
+  type: cogito
+  check_every: never
+  source:
+    sinks:
+      - gchat
+    gchat_webhook: ((gchat_webhook))
+```
+
 # Effects
 
 ## Build states mapping
@@ -158,7 +172,9 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
 
 # Source Configuration
 
-## Required keys
+## GitHub commit status only
+
+### Required keys
 
 - `owner`\
   The GitHub user or organization.
@@ -170,17 +186,34 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
   The OAuth access token.\
   See also: section [GitHub OAuth token](#github-oauth-token).
 
-## Optional keys
+### Optional keys
 
 - `context_prefix`\
   The prefix for the GitHub Commit status API "context" (see section [Effects on GitHub](#effects-on-github)). If present, the context will be set as `context_prefix/job_name`.\
   Default: empty.\
   See also: the optional `context` in the [put step](#the-put-step).
 
+- `log_level`:\
+  The log level (one of `debug`, `info`, `warn`, `error`, `silent`).\
+  Default: `info`.
+
+- `log_url`. **DEPRECATED, no-op, will be removed**\
+  A Google Hangout Chat webhook. Useful to obtain logging for the `check` step for Concourse < v7.x
+
+
+## GitHub commit status plus chat notifications
+
+### Required keys
+
+- The keys required for [Only GitHub commit status](##github-commit-status-only)
+
 - `gchat_webhook`\
   URL of a [Google Chat webhook]. A notification about the build status will be sent to the associated chat space, using a thread key composed by the pipeline name and commit hash.\
-  Default: empty.\
   See also: `chat_notify_on_states` and section [Effects on Google Chat](#effects-on-google-chat).
+
+### Optional keys
+
+- The optional keys for [Only GitHub commit status](##github-commit-status-only)
 
 - `chat_notify_on_states`\
   The build states that will cause a chat notification. One or more of `abort`, `error`, `failure`, `pending`, `success`.\
@@ -192,12 +225,29 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
   Default: `true`.\
   See also: the default build summary in [Effects on Google Chat](#effects-on-google-chat).
 
-- `log_level`:\
-  The log level (one of `debug`, `info`, `warn`, `error`, `silent`).\
-  Default: `info`.
+## Chat notification only
 
-- `log_url`. **DEPRECATED, no-op, will be removed**\
-  A Google Hangout Chat webhook. Useful to obtain logging for the `check` step for Concourse < v7.x
+### Required keys
+
+- `sinks`\
+  The sinks list for chat notification.
+  Acceptable values: `[gchat]`.
+
+- `gchat_webhook`\
+  URL of a [Google Chat webhook]. A notification will be sent to the associated chat space.\
+  See also: `chat_notify_on_states` and section [Effects on Google Chat](#effects-on-google-chat).
+
+### Optional keys
+
+- `chat_notify_on_states`\
+  The build states that will cause a chat notification. One or more of `abort`, `error`, `failure`, `pending`, `success`.\
+  Default: `[abort, error, failure]`.\
+  See also: section [Build states mapping](#build-states-mapping).
+
+- `chat_append_summary`\
+  One of: `true`, `false`. If `true`, append the default build summary to the custom `put.params.chat_message` and/or `put.params.chat_message_file`.\
+  Default: `true`.\
+  See also: the default build summary in [Effects on Google Chat](#effects-on-google-chat).
 
 ## Suggestions
 
@@ -232,6 +282,10 @@ If the `source` block has the optional key `gchat_webhook`, then it will also se
   See also: [Effects on GitHub](#effects-on-github), `source.context_prefix`.
 
 ## Optional params for chat
+
+- `sinks`\
+  If present, overrides `source.sinks`. This allows to run step `put` for example to send to chat only.
+  Default: `source.sinks`.
 
 - `gchat_webhook`\
   If present, overrides `source.gchat_webhook`. This allows to use the same Cogito resource for multiple chat spaces.\
@@ -270,6 +324,29 @@ on_success:
   params:
     state: success
     chat_message_file: the-message-dir/msg.txt
+```
+
+If using send to chat only and the `chat_message_file` parameter, the put step requires only one ["put inputs"]. For example:
+
+```yaml
+on_success:
+  put: chat-only
+  # the-message-dir: "output" of a previous task
+  inputs: [the-message-dir]
+  params:
+    state: success
+    chat_message_file: the-message-dir/msg.txt
+```
+
+If using send to chat only without `chat_message_file` parameter, `inputs` may be left empty.
+
+```yaml
+on_success:
+  put: gh-status
+  inputs: []
+  params:
+    state: success
+    chat_message: "This is the custom chat message."
 ```
 
 The reasons of this strictness is to help you have an efficient pipeline, since if the "put inputs" list is not set explicitly, then Concourse will stream all inputs used by the job to this resource, which can have a big performance impact. From the ["put inputs"] documentation:
--- a/cogito/check_test.go
+++ b/cogito/check_test.go
@@ -31,7 +31,7 @@ func TestCheckSuccess(t *testing.T) {
 		assert.DeepEqual(t, have, tc.wantOut)
 	}
 
-	baseSource := cogito.Source{
+	baseGithubSource := cogito.Source{
 		Owner:       "the-owner",
 		Repo:        "the-repo",
 		AccessToken: "the-token",
@@ -41,14 +41,14 @@ func TestCheckSuccess(t *testing.T) {
 		{
 			name: "first request (Concourse omits the version field)",
 			request: cogito.CheckRequest{
-				Source: baseSource,
+				Source: baseGithubSource,
 			},
 			wantOut: []cogito.Version{{Ref: "dummy"}},
 		},
 		{
 			name: "subsequent requests (Concourse adds the version field)",
 			request: cogito.CheckRequest{
-				Source:  baseSource,
+				Source:  baseGithubSource,
 				Version: cogito.Version{Ref: "dummy"},
 			},
 			wantOut: []cogito.Version{{Ref: "dummy"}},
@@ -80,7 +80,7 @@ func TestCheckFailure(t *testing.T) {
 		assert.Error(t, err, tc.wantErr)
 	}
 
-	baseSource := cogito.Source{
+	baseGithubSource := cogito.Source{
 		Owner:       "the-owner",
 		Repo:        "the-repo",
 		AccessToken: "the-token",
@@ -88,14 +88,30 @@ func TestCheckFailure(t *testing.T) {
 
 	testCases := []testCase{
 		{
-			name:    "validation failure: missing keys",
+			name:    "validation failure: missing repo keys",
 			source:  cogito.Source{},
 			writer:  io.Discard,
 			wantErr: "check: source: missing keys: owner, repo, access_token",
 		},
 		{
+			name: "validation failure: missing gchat keys",
+			source: cogito.Source{
+				Sinks: []string{"gchat"},
+			},
+			writer:  io.Discard,
+			wantErr: "check: source: missing keys: gchat_webhook",
+		},
+		{
+			name: "validation failure: wrong sink key",
+			source: cogito.Source{
+				Sinks: []string{"ghost", "gchat"},
+			},
+			writer:  io.Discard,
+			wantErr: "check: source: invalid sink(s): [ghost]",
+		},
+		{
 			name:    "write error",
-			source:  baseSource,
+			source:  baseGithubSource,
 			writer:  &testhelp.FailingWriter{},
 			wantErr: "check: preparing output: test write error",
 		},
--- a/cogito/gchatsink.go
+++ b/cogito/gchatsink.go
@@ -111,10 +111,6 @@ func gChatBuildSummaryText(gitRef string, state BuildState, src Source, env Envi
 	// https://github.com/Pix4D/cogito/commit/e8c6e2ac0318b5f0baa3f55
 	job := fmt.Sprintf("<%s|%s/%s>",
 		concourseBuildURL(env), env.BuildJobName, env.BuildName)
-	commitUrl := fmt.Sprintf("https://github.com/%s/%s/commit/%s",
-		src.Owner, src.Repo, gitRef)
-	commit := fmt.Sprintf("<%s|%.10s> (repo: %s/%s)",
-		commitUrl, gitRef, src.Owner, src.Repo)
 
 	// Unfortunately the font is proportional and doesn't support tabs,
 	// so we cannot align in columns.
@@ -123,7 +119,14 @@ func gChatBuildSummaryText(gitRef string, state BuildState, src Source, env Envi
 	fmt.Fprintf(&bld, "*pipeline* %s\n", env.BuildPipelineName)
 	fmt.Fprintf(&bld, "*job* %s\n", job)
 	fmt.Fprintf(&bld, "*state* %s\n", decorateState(state))
-	fmt.Fprintf(&bld, "*commit* %s\n", commit)
+	// An empty gitRef means that cogito has been configured as chat only.
+	if gitRef != "" {
+		commitUrl := fmt.Sprintf("https://github.com/%s/%s/commit/%s",
+			src.Owner, src.Repo, gitRef)
+		commit := fmt.Sprintf("<%s|%.10s> (repo: %s/%s)",
+			commitUrl, gitRef, src.Owner, src.Repo)
+		fmt.Fprintf(&bld, "*commit* %s\n", commit)
+	}
 
 	return bld.String()
 }
--- a/cogito/get_test.go
+++ b/cogito/get_test.go
@@ -31,7 +31,7 @@ func TestGetSuccess(t *testing.T) {
 		assert.DeepEqual(t, have, tc.wantOut)
 	}
 
-	baseSource := cogito.Source{
+	baseGithubSource := cogito.Source{
 		Owner:       "the-owner",
 		Repo:        "the-repo",
 		AccessToken: "the-token",
@@ -41,7 +41,7 @@ func TestGetSuccess(t *testing.T) {
 		{
 			name: "returns requested version",
 			request: cogito.GetRequest{
-				Source:  baseSource,
+				Source:  baseGithubSource,
 				Version: cogito.Version{Ref: "banana"},
 			},
 			wantOut: cogito.Output{Version: cogito.Version{Ref: "banana"}},
@@ -79,7 +79,7 @@ func TestGetFailure(t *testing.T) {
 		assert.Error(t, err, tc.wantErr)
 	}
 
-	baseSource := cogito.Source{
+	baseGithubSource := cogito.Source{
 		Owner:       "the-owner",
 		Repo:        "the-repo",
 		AccessToken: "the-token",
@@ -94,13 +94,13 @@ func TestGetFailure(t *testing.T) {
 		},
 		{
 			name:    "concourse validation failure: empty version field",
-			source:  baseSource,
+			source:  baseGithubSource,
 			writer:  io.Discard,
 			wantErr: "get: empty 'version' field",
 		},
 		{
 			name:    "concourse validation failure: missing output directory",
-			source:  baseSource,
+			source:  baseGithubSource,
 			version: cogito.Version{Ref: "dummy"},
 			writer:  io.Discard,
 			wantErr: "get: arguments: missing output directory",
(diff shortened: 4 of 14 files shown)


Esysc:
@Pix4D/integration the PR is ready for review

README.md
```diff
@@ -158,7 +172,9 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
 
 # Source Configuration
 
-## Required keys
+## GitHub commit status only
+
+### Required keys
```
aliculPix4D, line 177: Suggestion: `### Required keys (default behaviour)`

aliculPix4D: or maybe not since you already have: `GitHub commit status only`

README.md
```diff
@@ -170,18 +186,36 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
...
+
+- `log_url`. **DEPRECATED, no-op, will be removed**\
+  A Google Hangout Chat webhook. Useful to obtain logging for the `check` step for Concourse < v7.x
+
+
+## GitHub commit status plus chat notifications
+
+### Required keys
+
+- The keys required for [Only GitHub commit status](##github-commit-status-only)
```
aliculPix4D, line 208: ` The keys required for ... plus the following key/value` 

Maybe I would add also `plus...`

Esysc: It's personal taste, this is a list and you can follow each items . Adding `plus` seems a bit redundant to me.

cogito/put_test.go
```diff
@@ -212,20 +212,79 @@ func TestPutterLoadConfigurationInvalidParamsFailure(t *testing.T) {
...
+}`)
+	wantErr := `put: source: invalid sink(s): [pizza]. Supported sinks: [gchat github]`
+	putter := cogito.NewPutter("dummy-API", hclog.NewNullLogger())
+
+	err := putter.LoadConfiguration(in, nil)
+
+	assert.Error(t, err, wantErr)
+}
+
+func TestPutterLoadConfigurationUnknownSinkPutParams(t *testing.T) {
```
aliculPix4D, line 257: What is the difference with? `TestPutterLoadConfigurationInvalidSinksFailure`?

Same error message, only difference is in sinks name. Either I am missing something or this is the same test.

aliculPix4D (approved):
Pre-approving. I left some minor comments. Otherwise, LGTM

cogito/put_test.go
```diff
@@ -128,6 +131,29 @@ func TestPutterLoadConfigurationSuccess(t *testing.T) {
...
+		"sinks": ["gchat", "github"]
+	},
+	  "params": {"sinks": ["gchat"]}
+	}`)
+	putter := cogito.NewPutter("dummy-API", hclog.NewNullLogger())
+
+	err := putter.LoadConfiguration(in, nil)
+	assert.NilError(t, err)
+
+	if len(putter.Request.Params.Sinks) > 1 {
```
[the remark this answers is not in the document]
Esysc: The `sinks` content is already tested in other places, but I agree with you that it should clear here that we want to find `[gchat]`, I will change.

cogito/put_test.go
Esysc: Good catch, thanks. You are right they are testing the same.

README.md
marco-m-pix4d: Exactly, Andrea revamped the README adding one subsection per usage type of Cogito (only GitHub, GitHub + chat, only chat). We felt that this was the best way to help the user understand how to use Cogito depending on their goal.

README.md
```diff
@@ -216,6 +269,7 @@ No-op.
 Sets the GitHub commit status for a given commit, following the [GitHub Commit status API].
 The same commit can have multiple statuses, differentiated by parameter `context`.
 
+Depending.......
```
Esysc, line 272: This is a left over, I will remove

cogito/putter.go
```diff
@@ -51,22 +51,40 @@ func (putter *ProdPutter) LoadConfiguration(input []byte, args []string) error {
...
+	// Validate sinks configuration if present, then decide.
+	_, err = putter.Sinks()
+	if err != nil {
+		return fmt.Errorf("put: arguments: %w", err)
+	}
+	sinks := putter.Request.Source.Sinks
+	if len(putter.Request.Params.Sinks) > 0 {
+		sinks = putter.Request.Params.Sinks
+	}
+	sinksSet := sets.From(sinks...)
```
marco-m-pix4d, line 64: Here also I think we can avoid the `Set` suffix and keep the code simpler:

```go
	var sinks *sets.Set[string]
	// if put.params.sinks is not empty it overrides source.params.sinks.
	if len(putter.Request.Params.Sinks) > 0 {
		sinks = sets.From(putter.Request.Params.Sinks...)
	} else {
		sinks = sets.From(putter.Request.Source.Sinks...)
	}
```

cogito/putter.go
```diff
@@ -51,22 +51,40 @@ func (putter *ProdPutter) LoadConfiguration(input []byte, args []string) error {
...
 	putter.log.Debug("", "state", buildState)
 
 	return nil
 }
 
 func (putter *ProdPutter) ProcessInputDir() error {
-	// putter.InputDir, corresponding to key "put:inputs:", should contain 1 or 2 dirs.
-	// If it contains one, we support autodiscovery by not requiring to name it, we know
+	// putter.InputDir, corresponding to key "put:inputs:", should contain 0, 1 or 2 dirs.
+	// If it contains zero, aka undefined, Cogito will only send to chat.
```
marco-m-pix4d, line 85: Note that:
- string `putter.InputDir` is _never_ "undefined" or empty, it is guaranteed to be the path to an existing directory.
- Such existing directory can be empty.
- What can be "undefined" is the "put inputs" configuration key.
- If the "put inputs" configuration key is "undefined", it does NOT imply that putter.InputDir has zero directories: this is the whole point: it could have multiple directories, none of them useful for cogito. This is the famous case where we return an error to help avoiding useless streaming of volumes.

cogito/putter.go
```diff
@@ -81,14 +99,39 @@ func (putter *ProdPutter) ProcessInputDir() error {
...
 	source := putter.Request.Source
+
+	sinks := sets.From(source.Sinks...)
+	if len(params.Sinks) > 0 {
+		sinks = sets.From(params.Sinks...)
+	}
+
 	var msgDir string
 
+	if len(putter.InputDir) == 0 && (sinks.Size() > 0 && !sinks.Contains("github")) {
```
marco-m-pix4d, line 110: Is it possible to come out with a more readable approach here?

marco-m-pix4d:
@Esysc Note that GitHub decided to hide part of the comments; be careful not to loose them :-)

pipelines/cogito-acceptance.yml
```diff
@@ -47,6 +47,16 @@ resources:
       gchat_webhook: ((gchat_webhook))
       chat_notify_on_states: [abort, error, failure, pending, success]
 
+  - name: cogito-notify-only
```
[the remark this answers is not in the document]
Esysc: >Remember to update also the Taskfile accordingly

This is the resource, the job name doesn't change, or am I missing something?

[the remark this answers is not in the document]
Esysc: I see, below I follow your suggestion and removed the `default` suffix, so updated the Taskfile as well.

cogito/check_test.go
```diff
@@ -31,28 +31,39 @@ func TestCheckSuccess(t *testing.T) {
...
 			},
 			wantOut: []cogito.Version{{Ref: "dummy"}},
 		},
+		{
+			name: "first request (Concourse omits the version field) gchat only",
+			request: cogito.CheckRequest{
+				Source: baseGchatSource,
+			},
+			wantOut: []cogito.Version{{Ref: "dummy"}},
+		},
```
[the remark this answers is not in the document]
Esysc: Right, it's useless and I removed.

cogito/gchatsink.go
```diff
@@ -123,7 +119,14 @@ func gChatBuildSummaryText(gitRef string, state BuildState, src Source, env Envi
 	fmt.Fprintf(&bld, "*pipeline* %s\n", env.BuildPipelineName)
 	fmt.Fprintf(&bld, "*job* %s\n", job)
 	fmt.Fprintf(&bld, "*state* %s\n", decorateState(state))
-	fmt.Fprintf(&bld, "*commit* %s\n", commit)
+	// If it is chat only, don't report not existing GitHub information.
```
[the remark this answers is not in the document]
Esysc: Right, your suggestion make sense.

cogito/get_test.go
```diff
@@ -31,17 +31,29 @@ func TestGetSuccess(t *testing.T) {
...
 		{
 			name: "returns requested version",
 			request: cogito.GetRequest{
-				Source:  baseSource,
+				Source:  baseGithubSource,
+				Version: cogito.Version{Ref: "banana"},
+			},
+			wantOut: cogito.Output{Version: cogito.Version{Ref: "banana"}},
+		},
+		{
```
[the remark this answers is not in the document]
Esysc: The test coverage doesn't change (as before).

cogito/protocol.go
```diff
@@ -212,18 +216,43 @@ func (src *Source) UnmarshalJSON(data []byte) error {
...
-	// Validate mandatory fields.
+	// Evaluate mandatory fields.
 	//
 	var mandatory []string
-	if src.Owner == "" {
-		mandatory = append(mandatory, "owner")
+
+	defaultSinks := []string{"gchat", "github"}
+	defaultSinksSet := sets.From(defaultSinks...)
+	sinksSet := sets.From(src.Sinks...)
```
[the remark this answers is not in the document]
Esysc: Correct!

cogito/protocol.go
```diff
@@ -212,18 +216,43 @@ func (src *Source) UnmarshalJSON(data []byte) error {
 // Validate verifies the Source configuration and applies defaults.
 func (src *Source) Validate() error {
 	//
-	// Validate mandatory fields.
```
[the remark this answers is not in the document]
Esysc: I was thinking about, the concept of `mandatory` and `optional` is now dynamic and my first idea was to change validate into -> evaluate. I restored the original.

cogito/protocol.go
```diff
@@ -212,18 +216,43 @@ func (src *Source) UnmarshalJSON(data []byte) error {
...
+		sinksNotValid := sinksSet.Difference(defaultSinksSet)
+		if sinksNotValid.Size() > 0 {
+			return fmt.Errorf("source: invalid sink(s): %s. Supported sinks: %s", sinksNotValid, defaultSinks)
+		}
 	}
-	if src.Repo == "" {
-		mandatory = append(mandatory, "repo")
+
+	if sinksSet.Size() == 0 || sinksSet.Contains("github") {
+		// No sinks implies backward compatibility mode where github is mandatory and gchat optional.
```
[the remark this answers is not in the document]
Esysc: Updated!

cogito/protocol_test.go
```diff
@@ -71,10 +71,41 @@ func TestSourceValidationFailure(t *testing.T) {
...
+		},
+		{
+			name: "multiple invalid sink source keys",
+			source: cogito.Source{
+				Sinks: []string{"coffee", "shop"},
+			},
+			wantErr: "source: invalid sink(s): [coffee shop]. Supported sinks: [gchat github]",
+		},
+		{
+			name: "multiple mixed invalid/valid sink source keys",
```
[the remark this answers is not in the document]
Esysc: I introduced this test to ensure that even with a big `sinks` list, only the unsupported items are reported. It doesn't increase the coverage, but it was meaningful for me to test that the code works as intended. I'm open to remove it.

cogito/protocol_test.go
```diff
@@ -71,10 +71,41 @@ func TestSourceValidationFailure(t *testing.T) {
...
+		{
+			name:    "missing mandatory gchat source keys",
+			source:  cogito.Source{Sinks: []string{"gchat"}},
+			wantErr: "source: missing keys: gchat_webhook",
+		},
+		{
+			name: "invalid sink source keys",
+			source: cogito.Source{
+				Sinks:        []string{"gchat", "ghost"},
+				GChatWebHook: "sensitive-gchat-webhook",
```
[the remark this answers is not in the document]
Esysc: Do you mean the line 87? No, the invalid sink will be intercepted first, so let me remove it.

cogito/putter.go
Esysc: Done, thanks

cogito/putter.go
Esysc: Yes, sure. no need to check the size of sinks.

cogito/putter.go
Esysc: Fixed!

Esysc:
@marco-m-pix4d I should have addressed all your remarks.

**Note on `Sinker()`**: I've moved the check to a private function, so now it doesn't return an error anymore and the code test coverage is now increased of `0.1%` (from `99.0 %` to `99.1 %`). The difference with master is still the fact that the function need to filter the  `[]Sinker` returned based on the optional `sinks` configuration.

README.md
```diff
@@ -170,18 +186,36 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
...
+## GitHub commit status plus chat notifications
+
+### Required keys
+
+- The keys required for [Only GitHub commit status](##github-commit-status-only)
+
 - `gchat_webhook`\
   URL of a [Google Chat webhook]. A notification about the build status will be sent to the associated chat space, using a thread key composed by the pipeline name and commit hash.\
   Default: empty.\
   See also: `chat_notify_on_states` and section [Effects on Google Chat](#effects-on-google-chat).
```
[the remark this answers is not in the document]
Esysc: Yes, it is a copy/paste of the original item, but it is optional in current master.

cogito/putter.go
```diff
@@ -185,6 +248,16 @@ func collectInputDirs(dir string) ([]string, error) {
...
+// validateSinks return an error if the user set an unsupported sink in source or put.params.
+func validateSinks(sinks *sets.Set[string]) error {
+	supportedSinks := sets.From("gchat", "github")
+	difference := sinks.Difference(supportedSinks)
+	if difference.Size() > 0 {
+		return fmt.Errorf("%s", difference)
+	}
+	return nil
+}
+
```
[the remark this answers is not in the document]
Esysc: Yes sure, I don't like either having to maintain different places for this thing.

cogito/putter.go
```diff
@@ -51,22 +51,42 @@ func (putter *ProdPutter) LoadConfiguration(input []byte, args []string) error {
...
+	if len(args) == 0 && required {
 		return fmt.Errorf("put: arguments: missing input directory")
 	}
-	putter.InputDir = args[0]
-	putter.log.Debug("", "input-directory", putter.InputDir)
-
+	if len(args) > 0 {
+		putter.InputDir = args[0]
+		putter.log.Debug("", "input-directory", putter.InputDir)
+	}
```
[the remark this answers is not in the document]
Esysc: >Said otherwise, having len(args) == 0 is a concourse resource protocol violation error

This a part of the code that I didn't touch, but I agree with you, I'll update.

cogito/putter.go
```diff
@@ -81,14 +101,38 @@ func (putter *ProdPutter) ProcessInputDir() error {
...
 	source := putter.Request.Source
+
+	sinks := sets.From(source.Sinks...)
+	if len(params.Sinks) > 0 {
+		sinks = sets.From(params.Sinks...)
+	}
+
 	var msgDir string
 
+	if len(putter.InputDir) == 0 && !sinks.Contains("github") {
```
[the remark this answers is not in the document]
Esysc: Right, 5 eyes is better than 2.

cogito/putter.go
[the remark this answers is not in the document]
marco-m-pix4d: I think that `required` has been added in this PR. In any case, good simplification, thanks Olivier.

Esysc:
@odormond thank you for your review. It gave me the occasion to investigate further, simplify  and also catch a bug  processing input dirs. I've removed some useless and wrong tests that the previous code was hiding. I have added  the `only-msgdir` directory in `testdata` that was missing.

Esysc:
@Pix4D/integration ready for another review.

marco-m-pix4d:
@Esysc I saw a lot of activity on the PR but it is not clear to me if all the points have been addressed; could you please mention the team when the PR is ready for review again? Merci :-)

cogito/putter.go
```diff
@@ -135,21 +178,41 @@ func (putter *ProdPutter) ProcessInputDir() error {
...
+
+	// Sinks configured in Put params get priority.
+	sinksParams := putter.Request.Source.Sinks
+	sinksParamsPut := putter.Request.Params.Sinks
+	if len(sinksParamsPut) > 0 {
+		sinksParams = sinksParamsPut
+	}
+	if len(sinksParams) == 0 {
+		// Default to all supported sinks.
+		sinksParams = []string{"github", "gchat"}
```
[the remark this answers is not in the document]
Esysc: I have reduced the places, now a function returns the slice of sinks and it is the single source point.

odormond (commented):
A good step forward. Thanks.

Unfortunately, it's still buggy.

cogito/putter.go
```diff
@@ -105,51 +130,72 @@ func (putter *ProdPutter) ProcessInputDir() error {
...
 	}
+
+	sinksParams := supportedSinks()
+	// Sinks configured in Put params get priority.
+	if len(putter.Request.Params.Sinks) > 0 {
+		sinksParams = putter.Request.Params.Sinks
+	}
+	if len(putter.Request.Source.Sinks) > 0 {
+		sinksParams = putter.Request.Source.Sinks
+	}
```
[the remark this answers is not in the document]
Esysc: Unfortunately is not that simple, but I understand the simplification that should introduce. It needs some rework, I think I could create a ticket for that, because it'd introduce more code changes.
My proposal is addressing the parameters priority bug, also add a test for that, I realized, is missing.

cogito/putter.go
```diff
@@ -105,51 +130,72 @@ func (putter *ProdPutter) ProcessInputDir() error {
...
+		// If the size is 0 after removing the directory containing the chat message
+		// and Cogito should update the commit status, return an error.
+		if sinks.Contains("github") {
+			return fmt.Errorf(
+				"put:inputs: missing directory for GitHub repo: have: %v, GitHub: %s/%s",
+				inputDirs, source.Owner, source.Repo)
+		}
+		putter.log.Debug("", "inputDirs", inputDirs, "msgDir", msgDir)
+	case 1:
+		repoDir := filepath.Join(putter.InputDir, inputDirs.OrderedList()[0])
```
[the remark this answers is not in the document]
marco-m-pix4d: How would you envision a different API? We own the `sets` implementation.

cogito/putter.go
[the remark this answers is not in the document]
marco-m-pix4d: > Make `Sinks` a private function

`Sinks` is part of the interface API

Esysc:
@Pix4D/integration I've finally terminated, it took some time to validate the code. The last commit adds a unit test to increase the coverage from `99.3%` to `99.5%` as it is in the current master.

README.md
```diff
@@ -192,12 +225,30 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
...
-  Default: `info`.
+## Chat notification only
 
-- `log_url`. **DEPRECATED, no-op, will be removed**\
-  A Google Hangout Chat webhook. Useful to obtain logging for the `check` step for Concourse < v7.x
+### Required keys
+
+- `sinks`\
+  The sinks for chat notification.
+  Default: `[]`.
```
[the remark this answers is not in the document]
Esysc: Fixed

cogito/protocol.go
```diff
@@ -215,15 +219,37 @@ func (src *Source) Validate() error {
 	// Validate mandatory fields.
 	//
 	var mandatory []string
-	if src.Owner == "" {
-		mandatory = append(mandatory, "owner")
+
+	// This is the Source validation about custom sinks,
+	// the second argument represents the put params that is nil
+	// at this stage.
+	_, err := MergeAndValidateSinks(src.Sinks, nil)
```
odormond, line 226: I really don't like the idea of merging the sinks, throwing the result away and then  restart from scratch with the argument of the merge.
If this newly introduced function returned the merged sinks as a `Set` instead of a list of strings, you could use it straight away.
```diff
-	_, err := MergeAndValidateSinks(src.Sinks, nil)
+	sinks, err := MergeAndValidateSinks(src.Sinks, nil)
	if err != nil {
		return fmt.Errorf("source: invalid sink(s): %w", err)
	}

-	sinks := sets.From(src.Sinks...)
	if sinks.Size() == 0 || sinks.Contains("github") {
```

cogito/putter.go
```diff
@@ -105,51 +122,63 @@ func (putter *ProdPutter) ProcessInputDir() error {
...
-	if inputDirs.Size() == 0 {
-		return fmt.Errorf(
-			"put:inputs: missing directory for GitHub repo: have: %v, GitHub: %s/%s",
-			collected, source.Owner, source.Repo)
-	} else if inputDirs.Size() > 1 {
+	switch inputDirs.Size() {
+	case 0:
+		// If the size is 0 after removing the directory containing the chat message
+		// and Cogito should update the commit status, return an error.
+		if sets.From(sinks...).Contains("github") {
```
odormond, line 129: As I said above, if it returned a Set it could be used directly.
```suggestion
		if sinks.Contains("github") {
```

cogito/putter.go
```diff
@@ -105,51 +122,63 @@ func (putter *ProdPutter) ProcessInputDir() error {
...
 			GitRef:   putter.gitRef,
 			Request:  putter.Request,
 		},
 	}
+	source := putter.Request.Source.Sinks
+	params := putter.Request.Params.Sinks
+	sinks, _ := MergeAndValidateSinks(source, params)
+
+	sinkers := make([]Sinker, 0, len(sinks))
+	for _, s := range sinks {
```
odormond, line 177: With `MergeAndValidateSinks` returning a Set this would just becomes:
```suggestion
	sinkers := make([]Sinker, 0, sinks.Size())
	for _, s := range sinks.OrderedList() {
```

cogito/putter.go
```diff
@@ -81,6 +91,13 @@ func (putter *ProdPutter) ProcessInputDir() error {
 
 	params := putter.Request.Params
 	source := putter.Request.Source
+
+	// Validate optional sinks configuration.
+	sinks, err := MergeAndValidateSinks(source.Sinks, params.Sinks)
```
odormond, line 96: Why the hell do we have to recompute, revalidate and rehandle the error once more in `ProcessInputDir` ?! All this already took place in `LoadConfiguration` and it is always called before `ProcessInputDir` and nothing could have changed in between. Is that just to avoid saving the merged sinks in `putter`? Why not?

Esysc: Because it doesn't work that way. As I said during the stand-up, I spent almost the whole day trying to figure out why `putter.Sinkers = putter.Sinks()` ended having the sinkers always empty breaking all the code. I really tried to have this object set just once and then available, but it didn't work.

cogito/protocol.go
Esysc: Not exactly, this is the source validation checking what is mandatory and what is optional.
We are not interested to get the sinks (defaults or overridden), so the first step is to check if there is any error giving `src.Sinks` as argument. I don't want to have all the supported sinks if there is nothing configured, that's why I'm throwing away `_` the sinks returned by the function, but only any eventual error.

odormond: You are interested in using what you validated and you return what you validated. So you it.

However, your comment made me realize that the code can be cleaned up further when using the returned sinks. The `if sinks.Size() == 0 || sinks.Contains("github") {` no longer needs to check if it's empty because what was returned is always a non-empty set of sinks. So you can just do `if sinks.Contains("github") {`. That's an excellent cleanup as it removes the needs to know about the default's logic here too.

cogito/putter.go
Esysc: fixed

cogito/putter.go
Esysc: fixed

cogito/protocol.go
Esysc: > You are interested in using what you validated and you return what you validated. So you it.
> 
> However, your comment made me realize that the code can be cleaned up further when using the returned sinks. The `if sinks.Size() == 0 || sinks.Contains("github") {` no longer needs to check if it's empty because what was returned is always a non-empty set of sinks. So you can just do `if sinks.Contains("github") {`. That's an excellent cleanup as it removes the needs to know about the default's logic here too.

ah, I get your point, let me do a try.

Esysc: I'm sorry it cannot work. The scope of `Validate()`, it's just to raise earlier if the resource is misconfigured (missing mandatory parameter. If I get all the default sinks at this stage, all the parameters become mandatory (I've 27 tests failing). 
We should consider to refactor this logic, but it's outside this PR. 
Also `Validate()` runs for the `check` and the `get` , there are some units tests around this.

cogito/putter.go
odormond: So I finally gave it a try to understand what you were talking about. Basically, in itself, it just works as I expected.
The actual issue is with the tests. Both `ProcessInputDirs` and `Sinks` now require a bit more state to be setup as `putter.sinks` is set by `LoadConfiguration`.
I just pushed a branch on top of yours that does just that: `refactor_putter_tests`. Might not be the best implementation however. So to be discussed. @marco-m-pix4d, can you have a look please?

Esysc: Meanwhile I removed the check in `ProcessInputDir()`, we want just the sinks there.

marco-m-pix4d: @odormond, before I take a look at the branch `refactor_putter_tests` (which might solve everything for the better for what I know), a more high level summary of the pair programming that Andrea and myself did Wednesday is the following:

- Yes, there is code duplication.
- On the other hand, the concern I had with your proposal is that it puts more state in the Putter object for the sake of reducing code duplication. This is a trade-off that is not always worth it: as you mention when referring to `refactor_putter_tests`, it requires to add more state in the tests.
- On the other hand, I am deep into Consul and I find it difficult, in these days, to put myself with enough depth in this code to follow all changes, so maybe I am making a mountain out of a mole hill.
- I think it is quite important to merge this PR so that we start having the feature and see how it behaves "for real". As long as we do not cut a release, nobody will pick it up in a pipeline in any case.

Would it be OK for you to proceed with merging this PR, and consider branch `refactor_putter_tests` as a follow-up PR? Or am I missing something that fundamental that we want to do only one merge?

cogito/putter.go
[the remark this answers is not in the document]
odormond: Even with all its batteries, python would only let you do a `pop()` (on both a set and a dict) so if the underlying map type support that would be an option.

odormond (approved):
Thanks for switching `MergeAndValidateSinks` to returning a set!

Let's go with this for now. We can still refactor further but as that will require significant changes to the tests too it's better left for another PR.

Thanks again for your efforts! I understand what kind of challenge it was.

marco-m-pix4d (approved):
@Esysc 

Let's do the following:
- merge this PR
- start using the generated docker image (with tag `master`) in our own pipelines to battle-test it, especially the independent chat send feature. NOTE: given the bug with Concourse at picking up a new version of the image, we must really ensure (by looking at the debug output in a pipeline) that we are actually picking up the new version. In case of doubts, just ask in our channel, we even have the SQL script to cleanup if needed.
- if all goes well, after a few days we can make a release.

Thanks for the effort.

Esysc:
Sorry for force-push, I rewrote the history.

PR #117, opened 2023-04-20
PCI-3118 github package refactoring

aliculPix4D:
**commit 1**: github: pass the sink logger to github package
- enables logging from inside github package

**commit 2**:github: use 'http.Response.Header.Get(key)' to fetch the header

Advantages:
- http.Response.Header.Get(key) is an existing built-in method
- it handles canonicalization (go http server converts header keys into case-sensitive keys)
- caller doesn't need to check if the key exists or how many elements it has

Using `http.Response.Header.Get(key)` will be much more important when inspecting `rate limit` related headers.

### DIFF
--- a/github/commitstatus.go
+++ b/github/commitstatus.go
@@ -13,6 +13,8 @@ import (
 	"path"
 	"strings"
 	"time"
+
+	"github.com/hashicorp/go-hclog"
 )
 
 // StatusError is one of the possible errors returned by the github package.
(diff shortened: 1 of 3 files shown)


github/commitstatus.go
```diff
@@ -83,6 +87,7 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 		return fmt.Errorf("JSON encode: %w", err)
 	}
 
+	s.log.Info("making a Github REST API http request")
```
aliculPix4D, line 90: this is only to show the usage example, so that I can enable the feature and use it in the same commit....

It can be removed later or better say it will be replaced with a msg logging how many attempts we are doing...

Esysc (approved):

PR #118, opened 2023-04-20
PCI-3118 rate limit retry

aliculPix4D:
**Commit 1:** (NO CHANGE IN THE BEHAVIOUR) github: refactor code to allow adding retry logic
- collect and store response status code, body and some header info (about oauth and rate limit)

**Commit 2:**  github: implement retry mechanism when rate limited

Just a draft for now to collect a bit of feedback...

If desired, I can already activate retry also for status codes >500 in this PR

### DIFF
--- a/github/commitstatus.go
+++ b/github/commitstatus.go
@@ -32,6 +32,12 @@ func (e *StatusError) Error() string {
 // API is the GitHub API endpoint.
 const API = "https://api.github.com"
 
+// Maximum number of retries for the retryable http request
+const maxRetries = 3
+
+// Maximum time to wait before next attempt
+var MaxSleepTime = 15 * time.Minute
+
 type CommitStatus struct {
 	server  string
 	token   string
@@ -88,7 +94,6 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 		return fmt.Errorf("JSON encode: %w", err)
 	}
 
-	s.log.Info("making a Github REST API http request")
 	req, err := http.NewRequest(http.MethodPost, url, bytes.NewBuffer(reqBodyJSON))
 	if err != nil {
 		return fmt.Errorf("create http request: %w", err)
@@ -97,12 +102,20 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 	req.Header.Set("Accept", "application/vnd.github.v3+json")
 	req.Header.Set("Content-Type", "application/json")
 
-	resp, err := httpRequestDo(req)
-	if err != nil {
-		return err
-	}
+	for attempt := 0; attempt < maxRetries; attempt++ {
+		s.log.Info(fmt.Sprintf("Http request attempt: %d out of %d", attempt+1, maxRetries))
+		resp, err := httpRequestDo(req)
+		if err != nil {
+			return err
+		}
 
-	return s.checkStatus(resp, state, sha, url)
+		retry, timeToSleep := checkForRetry(resp, attempt)
+		if !retry {
+			return s.checkStatus(resp, state, sha, url)
+		}
+		time.Sleep(timeToSleep)
+	}
+	return nil
 }
 
 type httpResponse struct {
@@ -200,4 +213,23 @@ func min(a, b int) int {
 		return a
 	}
 	return b
+}
+
+// checkForRetry determines if we should retry the http request and caluclates wait time between retries
+func checkForRetry(res httpResponse, attempt int) (bool, time.Duration) {
+	switch {
+	case attempt == maxRetries-1:
+		return false, 0
+	// If you exceed the rate limit, the response will have a 403 status and the x-ratelimit-remaining header will be 0
+	// https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit
+	case res.statusCode == 403 && res.rateLimitRemaining == 0:
+		sleepTime := time.Until(res.rateLimitReset)
+		if sleepTime > MaxSleepTime {
+			return false, 0
+		} else {
+			return true, time.Until(res.rateLimitReset)
+		}
+	default:
+		return false, 0
+	}
 }
--- a/github/commitstatus.go
+++ b/github/commitstatus.go
@@ -36,7 +36,7 @@ const API = "https://api.github.com"
 const maxRetries = 3
 
 // Maximum sleep time allowed
-var maxSleepTime = 15 * time.Minute
+var MaxSleepTime = 15 * time.Minute
 
 type CommitStatus struct {
 	server  string
@@ -103,18 +103,17 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 	req.Header.Set("Content-Type", "application/json")
 
 	var response httpResponse
-	var retry bool
-	timeToSleep := time.Second * 0
+	timeToSleep := 0 * time.Second // 0 seconds
 
 	for attempt := 1; attempt <= maxRetries; attempt++ {
 		time.Sleep(timeToSleep)
-		s.log.Info(fmt.Sprintf("Http request attempt: %d out of %d", attempt, maxRetries))
+		s.log.Info("GitHub HTTP request", "attempt", attempt, "max", maxRetries)
 		response, err = httpRequestDo(req)
 		if err != nil {
 			return err
 		}
-		retry, timeToSleep = checkForRetry(response)
-		if !retry {
+		timeToSleep = checkForRetry(response)
+		if timeToSleep == 0 {
 			break
 		}
 	}
@@ -194,7 +193,7 @@ func (s CommitStatus) checkStatus(resp httpResponse, state, sha, url string) err
 		hint = "Either wrong credentials or PAT expired (check your email for expiration notice)"
 	case http.StatusForbidden:
 		if resp.rateLimitRemaining == 0 {
-			hint = "Rate limited but the wait time to reset would be longer than 15 minutes (MaxSleepTime)"
+			hint = fmt.Sprintf("Rate limited but the wait time to reset would be longer than %v (MaxSleepTime)", MaxSleepTime)
 		} else {
 			hint = "none"
 		}
@@ -225,19 +224,19 @@ func min(a, b int) int {
 	return b
 }
 
-// checkForRetry determines if we should retry the http request and caluclates wait time between retries
-func checkForRetry(res httpResponse) (bool, time.Duration) {
+// checkForRetry determines if we should retry the http request and calculates wait time between retries
+func checkForRetry(res httpResponse) time.Duration {
 	switch {
 	// If you exceed the rate limit, the response will have a 403 status and the x-ratelimit-remaining header will be 0
 	// https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit
 	case res.statusCode == http.StatusForbidden && res.rateLimitRemaining == 0:
 		sleepTime := time.Until(res.rateLimitReset)
-		if sleepTime > maxSleepTime {
-			return false, 0
+		if sleepTime > MaxSleepTime {
+			return 0
 		} else {
-			return true, sleepTime
+			return sleepTime
 		}
 	default:
-		return false, 0
+		return 0
 	}
 }
--- a/github/commitstatus.go
+++ b/github/commitstatus.go
@@ -192,6 +192,12 @@ func (s CommitStatus) checkStatus(resp httpResponse, state, sha, url string) err
 		hint = "Github API is down"
 	case http.StatusUnauthorized:
 		hint = "Either wrong credentials or PAT expired (check your email for expiration notice)"
+	case http.StatusForbidden:
+		if resp.rateLimitRemaining == 0 {
+			hint = "Rate limited but the wait time to reset would be longer than 15 minutes (MaxSleepTime)"
+		} else {
+			hint = "none"
+		}
 	default:
 		// Any other error
 		hint = "none"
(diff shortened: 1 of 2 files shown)


github/commitstatus.go
```diff
@@ -96,14 +102,44 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
...
+			return err
+		}
+
+		retry, timeToSleep := checkForRetry(resp, attempt)
+		if !retry {
+			return s.checkStatus(resp, state, sha, url)
+		}
+		time.Sleep(timeToSleep)
+	}
+	return nil
```
aliculPix4D, line 118: should I report an error here?

I don't know how this can be reached but if we reach it its an error....

github/commitstatus.go
```diff
@@ -96,14 +102,44 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
...
+
+type httpResponse struct {
+	statusCode         int
+	body               string
+	oauthInfo          string
+	rateLimitRemaining int
+	rateLimitReset     time.Time
+}
+
+func httpRequestDo(req *http.Request) (httpResponse, error) {
```
aliculPix4D, line 129: does the actual request, closes the response body (important), collects and stores all the data that would be required in subsequent steps...

github/commitstatus.go
```diff
@@ -172,3 +214,22 @@ func min(a, b int) int {
...
+	case attempt == maxRetries-1:
+		return false, 0
+	// If you exceed the rate limit, the response will have a 403 status and the x-ratelimit-remaining header will be 0
+	// https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit
+	case res.statusCode == 403 && res.rateLimitRemaining == 0:
+		sleepTime := time.Until(res.rateLimitReset)
+		if sleepTime > MaxSleepTime {
+			return false, 0
+		} else {
+			return true, time.Until(res.rateLimitReset)
```
aliculPix4D, line 230: TODO: I could reuse `sleepTime` here.

github/commitstatus.go
```diff
@@ -172,3 +214,22 @@ func min(a, b int) int {
...
+		}
+	default:
```
aliculPix4D, line 232:
```
// sleep for 5 seconds
case res.statusCode >= 500
   return true, time.Second*5
```

github/commitstatus_test.go
```diff
@@ -14,31 +15,59 @@ import (
...
 	}{
 		{
-			name:   "No errors",
-			body:   "Anything goes...",
-			status: http.StatusCreated,
+			name: "No errors",
+			response: mockedResponse{
+				body:               "Anything goes...",
+				statuses:           []int{http.StatusCreated},
+				rateLimitRemaining: []string{"5000"},
```
aliculPix4D, line 41: not really important in this test

github/commitstatus.go
```diff
@@ -31,6 +32,12 @@ func (e *StatusError) Error() string {
 // API is the GitHub API endpoint.
 const API = "https://api.github.com"
 
+// Maximum number of retries for the retryable http request
+const maxRetries = 3
+
+// Maximum time to wait before next attempt
```
odormond, line 38: The comment is wrong. Should we have to sleep for more than that we would not and just give up so there is no next attempt.

github/commitstatus.go
```diff
@@ -96,14 +102,44 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 	req.Header.Set("Accept", "application/vnd.github.v3+json")
 	req.Header.Set("Content-Type", "application/json")
 
+	for attempt := 0; attempt < maxRetries; attempt++ {
+		s.log.Info(fmt.Sprintf("Http request attempt: %d out of %d", attempt+1, maxRetries))
```
odormond, line 106: Note that instead of bothering with the `+1` and `-1` here and there, you could change the loop:
```suggestion
	for attempt := 1; attempt <= maxRetries; attempt++ {
		s.log.Info(fmt.Sprintf("Http request attempt: %d out of %d", attempt, maxRetries))
```

github/commitstatus.go
odormond: You should do something like this:
```suggestion
    s.log.Warn("Maximum number of retries reached. Giving up.")
	return s.checkStatus(resp, state, sha, url)
```

aliculPix4D: Are you sure, we will exit here if maximum retries is reached? My understanding is that we would still exit at L114 (`return s.checkStatus(resp, state, sha, url)`)

github/commitstatus.go
```diff
@@ -172,3 +214,22 @@ func min(a, b int) int {
 	}
 	return b
 }
+
+// checkForRetry determines if we should retry the http request and caluclates wait time between retries
+func checkForRetry(res httpResponse, attempt int) (bool, time.Duration) {
+	switch {
+	case attempt == maxRetries-1:
```
odormond, line 221: With the above loop modification:
```suggestion
	case attempt == maxRetries:
```

And now I see why you wondered about the `return nil` above.
I've not thought it trough but you can probably ignore that case here and let the loop terminate and handle the maxRetries as I described it above.

aliculPix4D: but the sleep is after... so we need to avoid sleeping before exiting the loop, which means I would just have to move this case as if else in the loop... which I didn't like...

github/commitstatus.go
aliculPix4D: Changed...

github/commitstatus.go
aliculPix4D: Changed as suggested...

github/commitstatus.go
aliculPix4D: > I've not thought it trough but you can probably ignore that case here and let the loop terminate and handle the maxRetries as I described it above.

See:
```
retry, timeToSleep := checkForRetry(resp, attempt)
if !retry {
   return s.checkStatus(resp, state, sha, url)
}
time.Sleep(timeToSleep)
```

In our discussion we said that `retry` should determine if we should retry or not. If max attempts is reached we should not retry. So this beahaviour is in accordance with what we agreed....

If on other hand I remove the case `case attempt == maxRetries:` we would say this is a retryable error, and we would reach: `time.Sleep(timeToSleep)`. Problem is that `timeToSleep` can be anything reported in the last request... so this is potentially problematic.. to avoid it I would need to introduce an if statement `if attempt == maxRetries` and return from this statement...
This is just reapeating `case attempt == maxRetries:` but in a wrong way...

github/commitstatus.go
```diff
@@ -31,6 +32,12 @@ func (e *StatusError) Error() string {
 // API is the GitHub API endpoint.
 const API = "https://api.github.com"
 
+// Maximum number of retries for the retryable http request
+const maxRetries = 3
```
Esysc, line 36: You are using lower camelcase here while the variable at line 39 is upper. Is there any reason about that?

aliculPix4D: Yes, tests are in different package so I wanted `MaxSleepTime` to be overridable from tests...

It turns out it was much simpler to achieve what I wanted:
```
// MaxSleepTime = 15 minutes (900s)
rateLimitReset: []int{now + 901},
```

Since the test I wanted to include did not really bring any value I removed it and did not include it in this PR. So to sumarize, yes I can rename already `MaxSleepTime` to `maxSleepTime` and if at some later time I need to override it from the tests, I will change what is needed then...

github/commitstatus.go
Esysc: ah I see the discussion later in this thread

aliculPix4D: After discussion with @odormond in private chat this issue was addressed in:
https://github.com/Pix4D/cogito/pull/118/commits/c8f7ab2d1fbd2f8ea898ee590f2500d2b7d5fe71

github/commitstatus.go
aliculPix4D: After discussion with @odormond in private chat this issue was addressed in:
https://github.com/Pix4D/cogito/pull/118/commits/c8f7ab2d1fbd2f8ea898ee590f2500d2b7d5fe71

github/commitstatus.go
aliculPix4D: renamed the other value....

github/commitstatus.go
```diff
@@ -172,3 +218,20 @@ func min(a, b int) int {
 	}
 	return b
 }
+
+// checkForRetry determines if we should retry the http request and caluclates wait time between retries
+func checkForRetry(res httpResponse) (bool, time.Duration) {
+	switch {
+	// If you exceed the rate limit, the response will have a 403 status and the x-ratelimit-remaining header will be 0
+	// https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit
+	case res.statusCode == 403 && res.rateLimitRemaining == 0:
```
aliculPix4D, line 227: replace with `case res.statusCode == http.StatusForbidden && res.rateLimitRemaining == 0:`?

odormond (approved):

github/commitstatus.go
```diff
@@ -172,3 +224,20 @@ func min(a, b int) int {
 	}
 	return b
 }
+
+// checkForRetry determines if we should retry the http request and caluclates wait time between retries
```
Esysc, line 228:
```suggestion
// checkForRetry determines if we should retry the http request and calculates wait time between retries
```

Esysc (approved):

aliculPix4D:
@Pix4D/integration I will put the PR on hold until I get the comments from @marco-m

github/commitstatus.go
```diff
@@ -96,14 +102,48 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 	req.Header.Set("Accept", "application/vnd.github.v3+json")
 	req.Header.Set("Content-Type", "application/json")
 
+	var response httpResponse
+	var retry bool
+	timeToSleep := time.Second * 0
```
marco-m-pix4d, line 107: The constants in `time` are to be considered as unit of measure, so it is customary to write the constant on the right of the multiplication:

```go
	timeToSleep := 0 * time.Second // 0 seconds
```

github/commitstatus.go
```diff
@@ -146,23 +192,29 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 		hint = "Github API is down"
 	case http.StatusUnauthorized:
 		hint = "Either wrong credentials or PAT expired (check your email for expiration notice)"
+	case http.StatusForbidden:
+		if resp.rateLimitRemaining == 0 {
+			hint = "Rate limited but the wait time to reset would be longer than 15 minutes (MaxSleepTime)"
```
marco-m-pix4d, line 197: 15 is hardcoded in the string and can become incorrect

github/commitstatus.go
marco-m-pix4d, line 228: typo caluclates

github/commitstatus.go
```diff
@@ -31,6 +32,12 @@ func (e *StatusError) Error() string {
 // API is the GitHub API endpoint.
 const API = "https://api.github.com"
 
+// Maximum number of retries for the retryable http request
+const maxRetries = 3
+
+// Maximum sleep time allowed
+var maxSleepTime = 15 * time.Minute
```
marco-m-pix4d, line 39: How did we come out with 15 minutes? My concern is that this will occupy a worker for 15 minutes, which can be a lot (for configurations like ours where we limit 1 task per worker). Or maybe the resource is not counted as a task ? I forgot.

My first reaction would be to use a smaller value, 5 or 10 ?

marco-m-pix4d:
@aliculPix4D I left some inline comments.

Regarding the approach I mentioned at the standup, it would involve injecting in the HTTP Client a custom Roundtripper (and this RoundTripper would take care of the retry). That to me would be the correct approach, on the other hand it is not as immediate as I thought, so it might be more pragmatic to use the current approach.

If I look at the code in the PR, my first impression is that it should be possible to simplify it, but I cannot put the finger on where.

github/commitstatus.go
aliculPix4D: I had a private discussion with @odormond about it.

The worker will not be occupied with cogito (resource), but it will be with pipeline setter (task).

The 15minutes because in most cases we recieved this error I noticed the rate reset time in between 10-15minutes. I was tempted to put 30min here actually. 

Remeber that build jobs can take sometimes even 1h, so if we put this value too small, I fear that this retry is usless.. 
On the other hand, we should not wait 1h to get a status check... As I wrote above, given that I usually see the rate reset time to require the waiting times up to 15minutes, I choose this value....

Personal opionion: implementing a retry for rate limit with wait time limited to 5min or 10min is way too short.

github/commitstatus.go
```diff
@@ -31,6 +32,12 @@ func (e *StatusError) Error() string {
 // API is the GitHub API endpoint.
 const API = "https://api.github.com"
 
+// Maximum number of retries for the retryable http request
+const maxRetries = 3
+
+// Maximum sleep time allowed
+var MaxSleepTime = 15 * time.Minute
```
aliculPix4D, line 39: I am exporing here because I used the var in tests which is a pkg with different name (black box testing)

github/commitstatus.go
```diff
@@ -96,14 +102,48 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 	req.Header.Set("Accept", "application/vnd.github.v3+json")
 	req.Header.Set("Content-Type", "application/json")
 
+	var response httpResponse
+	var retry bool
+	timeToSleep := time.Second * 0
+
+	for attempt := 1; attempt <= maxRetries; attempt++ {
+		time.Sleep(timeToSleep)
+		s.log.Info(fmt.Sprintf("Http request attempt: %d out of %d", attempt, maxRetries))
```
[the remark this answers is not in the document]

github/commitstatus.go

github/commitstatus.go

github/commitstatus.go
```diff
@@ -172,3 +224,20 @@ func min(a, b int) int {
 	}
 	return b
 }
+
+// checkForRetry determines if we should retry the http request and caluclates wait time between retries
+func checkForRetry(res httpResponse) (bool, time.Duration) {
```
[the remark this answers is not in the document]

aliculPix4D:
@marco-m-pix4d I addressed all the comments and give you my opinion on the value of `MaxSleepTime` variable.

marco-m-pix4d (approved):
Note that we need to mention this behavior (retry in case of rate limit) both in the README and CHANGELOG.

aliculPix4D:
> Note that we need to mention this behavior (retry in case of rate limit) both in the README and CHANGELOG.

Yes, will do that in another PR...

PR #119, opened 2023-04-25
github: retry on status codes >= 500

aliculPix4D:
github: retry on status codes >= 500

This will enable us to get rid of the Concourse retrys in developers pipelines. Remeber that developers added around 10 Concourse retrie around some steps i.e cogito put step. This was probably making the issue with rate limit even worse. On the other hand it was partially justified to not waste 1h build time just because of a transient github error while posting the status.

Real Concourse errors (worker disapered) are alredy handled by the Concourse itself, so after this is implemented we might be able to convince people to not use retry...

Example: https://concourse-ci.org/attempts-step.html
Example from Pix4D: https://github.com/Pix4D/pix4d-geospatial-desktop/blob/master/ci/pipelines/survey-template.yml#L7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v0.10.0] - 2023-XX-XX
+
+- Implements retry logic for Github API Http requests. Cogito retries the http request up to 3 times in case Github user is rate limited. The maximum wait time between request is set to 15 minutes. Additionally, all http requests with status code `>=500` (i.e `500 Internal Server Error`, `503 Service Unavailable`, `504 Gateway Timeout`) are retried. For status codes `>=500` the wait time between retries is set to 5 seconds.
+
 ## [v0.9.0] - 2023-03-16
 
-This release allows to use cogito purely to send chat notifications, completely independently from GitHub. This allows to use cogito in place of concourse-hangouts-resource in any situation.
+- This release allows to use cogito purely to send chat notifications, completely independently from GitHub. This allows to use cogito in place of concourse-hangouts-resource in any situation.
 
 ### Added
 
@@ -226,4 +230,4 @@ This release allows to use cogito for the vast majority of chat notifications wh
 [v0.8.1]: https://github.com/Pix4D/cogito/releases/tag/v0.8.1
 [v0.8.2]: https://github.com/Pix4D/cogito/releases/tag/v0.8.2
 [v0.9.0]: https://github.com/Pix4D/cogito/releases/tag/v0.9.0
-
+[v0.10.0]: https://github.com/Pix4D/cogito/releases/tag/v0.10.0
--- a/github/commitstatus.go
+++ b/github/commitstatus.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-hclog"
+	"golang.org/x/exp/slices"
 )
 
 // StatusError is one of the possible errors returned by the github package.
@@ -32,14 +33,19 @@ func (e *StatusError) Error() string {
 // API is the GitHub API endpoint.
 const API = "https://api.github.com"
 
-// Maximum number of retries for the retryable http request
-const maxRetries = 3
+type Target struct {
+	Server string
 
-// Maximum sleep time allowed
-var MaxSleepTime = 15 * time.Minute
+	// Maximum number of retries for the retryable http request
+	MaxRetries int
+	// Default wait time between two http requests
+	WaitTime time.Duration
+	// Maximum sleep time allowed
+	MaxSleepTime time.Duration
+}
 
 type CommitStatus struct {
-	server  string
+	target  Target
 	token   string
 	owner   string
 	repo    string
@@ -57,8 +63,8 @@ type CommitStatus struct {
 //
 // See also:
 // https://docs.github.com/en/rest/commits/statuses#about-the-commit-statuses-api
-func NewCommitStatus(server, token, owner, repo, context string, log hclog.Logger) CommitStatus {
-	return CommitStatus{server, token, owner, repo, context, log}
+func NewCommitStatus(target Target, token, owner, repo, context string, log hclog.Logger) CommitStatus {
+	return CommitStatus{target, token, owner, repo, context, log}
 }
 
 // AddRequest is the JSON object sent to the API.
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -5,9 +5,13 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v0.10.0] - 2023-04-26
+
+- Implements retry logic for Github API Http requests. Cogito retries the http request up to 3 times in case Github user is rate limited. The maximum wait time between request is set to 15 minutes. Additionally, all http requests with status code `>=500` (i.e `500 Internal Server Error`, `503 Service Unavailable`, `504 Gateway Timeout`) are retried. For status codes `>=500` the wait time between retries is set to 5 seconds.
+
 ## [v0.9.0] - 2023-03-16
 
-This release allows to use cogito purely to send chat notifications, completely independently from GitHub. This allows to use cogito in place of concourse-hangouts-resource in any situation.
+- This release allows to use cogito purely to send chat notifications, completely independently from GitHub. This allows to use cogito in place of concourse-hangouts-resource in any situation.
```
aliculPix4D, line 14: for consistency ...

github/commitstatus.go
```diff
@@ -225,17 +227,19 @@ func min(a, b int) int {
...
 	// https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit
 	case res.statusCode == http.StatusForbidden && res.rateLimitRemaining == 0:
 		sleepTime := time.Until(res.rateLimitReset)
-		if sleepTime > MaxSleepTime {
+		if sleepTime > maxSleepTime {
 			return 0
 		} else {
 			return sleepTime
 		}
+	case res.statusCode >= 500:
```
odormond, line 241: I just want to make it loud and clear, the basic rule is: do not retry on 5xx, unless you understand what you are doing.

The 5xx category of error is meant for cases when you, as a client, cannot know the server's state. So as a general rule one must assume that it is not safe to retry a request that failed with an error code in that category.

For example, take the typical 504 Gateway Timeout, it means the server took so long to process the request that the gateway gave up before seeing the answer. As a client you don't know if it processed it successfully or not. If you blindly retry it might happen that the request is successfully executed each time and if that request was to pay for something, you might end up paying N times instead of 1. But it might also fail on the second call, for instance because the transaction id was consumed in the previous (successful) call, and seeing the request failed the user might want to retry from the beginning thus purchasing something twice. This should show that it is always tricky to interpret 5xx and decide whether to retry or not unless you get a good understanding of what's going on.

Really keep in mind that whenever you want to retry an API call. Always ensure it is safe to do so for that particular API call.

In the case of cogito, it is safe to retry. At worst we'll be setting the commit status multiple times, which should be an idempotent operation anyway so that's safe.

github/commitstatus_test.go
```diff
@@ -56,6 +56,15 @@ func TestGitHubStatusSuccessMockAPI(t *testing.T) {
 				rateLimitReset: []int{now + 1, now + 3600},
 			},
 		},
+		{
+			name: "Github is down (Gateway timeout) in the first attempt, success in second attempt",
```
odormond, line 60:
```suggestion
			name: "Github is flaky (Gateway timeout) in the first attempt, success in second attempt",
```
In a sense, it's not really down if it succeeded on the second attempt.

github/commitstatus.go
aliculPix4D: > This risks becoming overengineering in the wrong place (the resource) to solve a human problem (people putting 10 retries) ...

Regarding the `overengineering` part. The following lines are the only required change....
```
case res.statusCode >= 500:
   return waitTime
```

marco-m-pix4d: In this case, the "overengineering" referred to the complexity in the behavior (this is a huge change), not to the complexity of the implementation.

github/commitstatus.go
```diff
@@ -57,8 +58,8 @@ type CommitStatus struct {
 //
 // See also:
 // https://docs.github.com/en/rest/commits/statuses#about-the-commit-statuses-api
-func NewCommitStatus(server, token, owner, repo, context string, log hclog.Logger) CommitStatus {
-	return CommitStatus{server, token, owner, repo, context, log}
+func NewCommitStatus(server, token, owner, repo, context string, log hclog.Logger, maxRetries int, waitTime, maxSleepTime time.Duration) CommitStatus {
```
marco-m-pix4d, line 61: The number of arguments to the function is now too much.
Ideally, we should pass a HTTP Client instead, capable of retrying, but this would require a custom RoundTripper as I already mentioned.

A possibility would be to collect everything related to HTTP in a struct: server, maxRetries, waitTime, maxSleepTime, and call it maybe target, so that we would have
```
func NewCommitStatus(target target, token, owner, repo, context string, log hclog.Logger)
```

github/commitstatus.go
```diff
@@ -105,17 +106,18 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
...
 		response, err = httpRequestDo(req)
 		if err != nil {
 			return err
 		}
-		timeToSleep = checkForRetry(response)
+		timeToSleep = checkForRetry(response, s.waitTime, s.maxSleepTime)
 		if timeToSleep == 0 {
 			break
 		}
+		s.log.Info("Sleeping for:", timeToSleep)
```
marco-m-pix4d, line 120: This will crash or report an error.
As I wrote in the previous comment, this is a structured logger and wants to be called as follows
```go
Info("this is the main message")
Info("this is the main message", "key1", value1, "key2", value2, ...)

```

github/commitstatus.go
marco-m-pix4d: Regarding the retry on 5xx, I thought a bit more. I think it is dangerous to simply detect >= 500, because some of the 5xx errors are permanent (see for example the list in https://en.wikipedia.org/wiki/List_of_HTTP_status_codes). So I think that a conservative list is 500,502,503,504. For these, we retry, For the others, we fail.

github/commitstatus_test.go
```diff
@@ -71,7 +80,7 @@ func TestGitHubStatusSuccessMockAPI(t *testing.T) {
 		defer ts.Close()
 
 		t.Run(tc.name, func(t *testing.T) {
-			ghStatus := github.NewCommitStatus(ts.URL, cfg.Token, cfg.Owner, cfg.Repo, context, hclog.NewNullLogger())
+			ghStatus := github.NewCommitStatus(ts.URL, cfg.Token, cfg.Owner, cfg.Repo, context, hclog.NewNullLogger(), 2, time.Second, time.Minute)
```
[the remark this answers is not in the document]
aliculPix4D: grouped in target...

github/commitstatus.go
aliculPix4D: Done as suggested.

github/commitstatus_test.go
aliculPix4D: Changed as suggested...

github/commitstatus.go

Esysc (approved):

marco-m-pix4d (approved):

PR #122, opened 2023-05-17
github: calculate sleep time based on Github server time

aliculPix4D:
- use date header reporting the server time to calculate the sleep time
- add a jitter time to calculated sleep time to prevent creating a Thundering Herd
- prevent negative sleep times. Minimal sleep time should be equal to jitter time

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v0.10.1] - 2023-05-17
+
+### Added
+
+- Adds a jitter time to calculated sleep time to prevent creating a Thundering Herd.
+
+### Fixed
+
+- Fixes a bug with negative times that could occur due to the small difference between local (client) time and Github server time. Sleep time is now calculated using the date header (time as seen by the Github server). Additionally, negative sleep times are prevented by setting a minimum sleep time equal to the jitter time.
+
 ## [v0.10.0] - 2023-04-28
+
+### Added
 
 - Implements retry logic for Github API HTTP requests. Cogito retries the HTTP request up to 3 times in case the Github user is rate limited. The maximum wait time between request is set to 15 minutes. Additionally, http requests with status codes 500 Internal Server Error, 502 Bad Gateway, 503 Service Unavailable, 504 Gateway Timeout are retried. For these 5xx status codes the wait time between retries is set to 5 seconds.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,15 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [v0.10.1] - 2023-05-17
+## [v0.10.1] - 2023-05-31
 
 ### Added
 
-- Adds a jitter time to calculated sleep time to prevent creating a Thundering Herd.
+- Prevent being rate limited by GitHub: add a jitter time to calculated sleep time to prevent creating a Thundering Herd.
 
 ### Fixed
 
-- Fixes a bug with negative times that could occur due to the small difference between local (client) time and Github server time. Sleep time is now calculated using the date header (time as seen by the Github server). Additionally, negative sleep times are prevented by setting a minimum sleep time equal to the jitter time.
+- Prevent being rate limited by GitHub: fix a bug that caused a negative sleep time due to the clock difference between GitHub and the host running Cogito.
 
 ## [v0.10.0] - 2023-04-28
 
--- a/cogito/ghcommitsink.go
+++ b/cogito/ghcommitsink.go
@@ -1,6 +1,7 @@
 package cogito
 
 import (
+	"math/rand"
 	"time"
 
 	"github.com/Pix4D/cogito/github"
@@ -38,6 +39,8 @@ func (sink GitHubCommitStatusSink) Send() error {
 		MaxRetries:   maxRetries,
 		WaitTime:     waitTime,
 		MaxSleepTime: maxSleepTime,
+		// adds some randomness to sleep time to prevent creating a Thundering Herd
+		Jitter: time.Duration(rand.Intn(30)) * time.Second,
 	}
 	commitStatus := github.NewCommitStatus(target, sink.Request.Source.AccessToken,
 		sink.Request.Source.Owner, sink.Request.Source.Repo, context, sink.Log)
--- a/cogito/ghcommitsink.go
+++ b/cogito/ghcommitsink.go
@@ -39,8 +39,7 @@ func (sink GitHubCommitStatusSink) Send() error {
 		MaxRetries:   maxRetries,
 		WaitTime:     waitTime,
 		MaxSleepTime: maxSleepTime,
-		// adds some randomness to sleep time to prevent creating a Thundering Herd
-		Jitter: time.Duration(rand.Intn(30)) * time.Second,
+		Jitter:       time.Duration(rand.Intn(30)) * time.Second,
 	}
 	commitStatus := github.NewCommitStatus(target, sink.Request.Source.AccessToken,
 		sink.Request.Source.Owner, sink.Request.Source.Repo, context, sink.Log)
--- a/github/commitstatus.go
+++ b/github/commitstatus.go
@@ -42,6 +42,8 @@ type Target struct {
 	WaitTime time.Duration
 	// Maximum sleep time allowed
 	MaxSleepTime time.Duration
+	// adds some randomness to sleep time to prevent creating a Thundering Herd
+	Jitter time.Duration
 }
 
 type CommitStatus struct {
@@ -107,6 +109,7 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 	req.Header.Set("Authorization", "token "+s.token)
 	req.Header.Set("Accept", "application/vnd.github.v3+json")
 	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("X-GitHub-Api-Version", "2022-11-28")
 
 	var response httpResponse
 	timeToSleep := 0 * time.Second // 0 seconds
@@ -118,7 +121,7 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 		if err != nil {
 			return err
 		}
-		timeToSleep = checkForRetry(response, s.target.WaitTime, s.target.MaxSleepTime)
+		timeToSleep = checkForRetry(response, s.target.WaitTime, s.target.MaxSleepTime, s.target.Jitter)
 		if timeToSleep == 0 {
 			break
 		}
@@ -134,6 +137,7 @@ type httpResponse struct {
 	oauthInfo          string
 	rateLimitRemaining int
 	rateLimitReset     time.Time
+	date               time.Time
 }
 
 func httpRequestDo(req *http.Request) (httpResponse, error) {
@@ -178,6 +182,12 @@ func httpRequestDo(req *http.Request) (httpResponse, error) {
 	response.rateLimitRemaining, _ = strconv.Atoi(resp.Header.Get("x-ratelimit-remaining"))
 	rateLimitReset, _ := strconv.Atoi(resp.Header.Get("x-ratelimit-reset"))
 	response.rateLimitReset = time.Unix(int64(rateLimitReset), 0)
+
+	// time.RFC1123 is the format of the Github date header
+	// example: Date:[ Mon, 02 Jan 2006 15:04:05 MST ]
+	date, _ := time.Parse(time.RFC1123, resp.Header.Get("date"))
+	response.date = date
+
 	return response, nil
 }
 
@@ -232,7 +242,7 @@ func min(a, b int) int {
 }
 
 // checkForRetry determines if we should retry the http request and calculates wait time between retries
-func checkForRetry(res httpResponse, waitTime, maxSleepTime time.Duration) time.Duration {
+func checkForRetry(res httpResponse, waitTime, maxSleepTime, jitter time.Duration) time.Duration {
 	retrayalbeStatusCodes := []int{
 		http.StatusInternalServerError, // 500
 		http.StatusBadGateway,          // 502
@@ -244,15 +254,23 @@ func checkForRetry(res httpResponse, waitTime, maxSleepTime time.Duration) time.
 	// If you exceed the rate limit, the response will have a 403 status and the x-ratelimit-remaining header will be 0
 	// https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit
 	case res.statusCode == http.StatusForbidden && res.rateLimitRemaining == 0:
-		sleepTime := time.Until(res.rateLimitReset)
+		sleepTime := res.rateLimitReset.Sub(res.date) + jitter
 		if sleepTime > maxSleepTime {
 			return 0
 		} else {
-			return sleepTime
+			// Don't allow negative sleep times. Minimum sleep time should be equal to jitter time.
+			return max(sleepTime, jitter)
 		}
 	case slices.Contains(retrayalbeStatusCodes, res.statusCode):
 		return waitTime
 	default:
 		return 0
 	}
+}
+
+func max(x, y time.Duration) time.Duration {
+	if x < y {
+		return y
+	}
+	return x
 }
--- a/github/commitstatus.go
+++ b/github/commitstatus.go
@@ -109,7 +109,6 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 	req.Header.Set("Authorization", "token "+s.token)
 	req.Header.Set("Accept", "application/vnd.github.v3+json")
 	req.Header.Set("Content-Type", "application/json")
-	req.Header.Set("X-GitHub-Api-Version", "2022-11-28")
 
 	var response httpResponse
 	timeToSleep := 0 * time.Second // 0 seconds
@@ -121,7 +120,10 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 		if err != nil {
 			return err
 		}
-		timeToSleep = checkForRetry(response, s.target.WaitTime, s.target.MaxSleepTime, s.target.Jitter)
+		timeToSleep, err = checkForRetry(response, s.target.WaitTime, s.target.MaxSleepTime, s.target.Jitter)
+		if err != nil {
+			return fmt.Errorf("internal error: %s", err)
+		}
 		if timeToSleep == 0 {
 			break
 		}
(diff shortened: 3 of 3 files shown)


github/commitstatus.go
```diff
@@ -244,15 +254,23 @@ func checkForRetry(res httpResponse, waitTime, maxSleepTime time.Duration) time.
...
 	// https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit
 	case res.statusCode == http.StatusForbidden && res.rateLimitRemaining == 0:
-		sleepTime := time.Until(res.rateLimitReset)
+		sleepTime := res.rateLimitReset.Sub(res.date) + jitter
 		if sleepTime > maxSleepTime {
 			return 0
 		} else {
-			return sleepTime
+			// Don't allow negative sleep times. Minimum sleep time should be equal to jitter time.
+			return max(sleepTime, jitter)
```
aliculPix4D, line 262: if we need to sleep, the minimal sleep time will a random number between 0-30.

github/commitstatus.go
```diff
@@ -178,6 +182,12 @@ func httpRequestDo(req *http.Request) (httpResponse, error) {
 	response.rateLimitRemaining, _ = strconv.Atoi(resp.Header.Get("x-ratelimit-remaining"))
 	rateLimitReset, _ := strconv.Atoi(resp.Header.Get("x-ratelimit-reset"))
 	response.rateLimitReset = time.Unix(int64(rateLimitReset), 0)
+
+	// time.RFC1123 is the format of the Github date header
+	// example: Date:[ Mon, 02 Jan 2006 15:04:05 MST ]
+	date, _ := time.Parse(time.RFC1123, resp.Header.Get("date"))
```
aliculPix4D, line 188: So the date header is always reported and its in `Mon, 02 Jan 2006 15:04:05 MST` format. I can't find anything about this header in Github documentation so I am worried with this a bit. Hence, on L112 I added `req.Header.Set("X-GitHub-Api-Version", "2022-11-28")` to at least try to ensure this doesn't start to break if there is some changes from Github side.

odormond: I'm not surprised you don't see it in the GitHub doc because this is a standard HTTP header. See https://datatracker.ietf.org/doc/html/rfc2616#section-14.18.

odormond (approved):

github/commitstatus.go
marco-m-pix4d: To add to Olivier's comment, if an header doesn't begin with "X-" AND the web server is well-behaved, for sure the header is a standard header and so is defined in an RFC.

If an header begins with "X-", then it depends. It can be something invented by the specific service (so in this case GtiHub) OR it could be still a standard header.

A very good starting point is https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

CHANGELOG.md
```diff
@@ -5,8 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v0.10.1] - 2023-05-17
+
+### Added
+
+- Adds a jitter time to calculated sleep time to prevent creating a Thundering Herd.
```
marco-m-pix4d, line 12: - Prevent being rate limited by GitHub: add ...

github/commitstatus.go
```diff
@@ -107,6 +109,7 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 	req.Header.Set("Authorization", "token "+s.token)
 	req.Header.Set("Accept", "application/vnd.github.v3+json")
 	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("X-GitHub-Api-Version", "2022-11-28")
```
marco-m-pix4d, line 112: per the review comments, this is not needed.

github/commitstatus.go
```diff
@@ -178,6 +182,12 @@ func httpRequestDo(req *http.Request) (httpResponse, error) {
 	response.rateLimitRemaining, _ = strconv.Atoi(resp.Header.Get("x-ratelimit-remaining"))
 	rateLimitReset, _ := strconv.Atoi(resp.Header.Get("x-ratelimit-reset"))
 	response.rateLimitReset = time.Unix(int64(rateLimitReset), 0)
+
+	// time.RFC1123 is the format of the Github date header
```
marco-m-pix4d, line 186: time.RFC1123 is the format of the HTTP `Date` header

github/commitstatus.go
```diff
@@ -244,15 +254,23 @@ func checkForRetry(res httpResponse, waitTime, maxSleepTime time.Duration) time.
 	// If you exceed the rate limit, the response will have a 403 status and the x-ratelimit-remaining header will be 0
 	// https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit
 	case res.statusCode == http.StatusForbidden && res.rateLimitRemaining == 0:
-		sleepTime := time.Until(res.rateLimitReset)
+		sleepTime := res.rateLimitReset.Sub(res.date) + jitter
 		if sleepTime > maxSleepTime {
 			return 0
 		} else {
-			return sleepTime
+			// Don't allow negative sleep times. Minimum sleep time should be equal to jitter time.
```
marco-m-pix4d, line 261: The comment is somewhat misleading:
We avoid negative sleep time only by using the server's notion of clock, the jitter is unrelated, but the comment is about the max, which is unrelated to negative values, because sleepTime is guaranteed to be non negative.

So I would remove the comment entirely.

marco-m-pix4d: See the comment above on how to make it clearer what is happening

github/commitstatus.go
```diff
@@ -244,15 +254,23 @@ func checkForRetry(res httpResponse, waitTime, maxSleepTime time.Duration) time.
 	// If you exceed the rate limit, the response will have a 403 status and the x-ratelimit-remaining header will be 0
 	// https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit
 	case res.statusCode == http.StatusForbidden && res.rateLimitRemaining == 0:
-		sleepTime := time.Until(res.rateLimitReset)
+		sleepTime := res.rateLimitReset.Sub(res.date) + jitter
```
marco-m-pix4d, line 257: Here we are conflating two different operations in one line and we are missing a comment explaining what is happening. We could do instead (untested):

```go
// Calculate sleeptime based solely on the server clock. This is unaffected by the clock drift between server and client.
sleepTime := res.rateLimitReset.Sub(res.date)
// Be a good netizen by adding some jitter to the time we sleep.
sleepTime := res.rateLimitReset.Sub(res.date) + jitter
```

github/commitstatus.go
marco-m-pix4d: I now start to think that this line is wrong. Above we just added jitter to sleepTime, here there should be no need to calculate any max, we should just return sleepTime...

marco-m-pix4d:
@aliculPix4D Let me know when you have gone through all the comments; if anything unclear, we can also have a chat if you prefer.

github/commitstatus.go
aliculPix4D: Yes, I agree with you but I wanted to ensure that no negative time can happen anymore.

So my thinking was that this line:
```
sleepTime := res.rateLimitReset.Sub(res.date) 
```
can still return anything if there is any change on the server side. So adding 30sec to something negative and larger than 30sec can still create a negative sleeptime.

It should never happend of course, and `res.rateLimitReset.Sub(res.date) ` should really always be a positive value, but I was sceptical... 

So I wanted to ensure we really can't have the negative time anymore...hence `max(sleeptime, jitter)`

I am fine with removing it, just wanted to make it clear that:
- I did this on purpose
- mostly in "panic", "paranoid" mode trying to ensure that we don't have negative sleep times anymore

marco-m-pix4d: I see. The problem with that approach is that the intent was hidden, so a reader is left thinking that the code is wrong. Now that I know the intent, I can advice.

From my understanding, the scenario you are mentioning can happen only in case of a clock severely misbehaving either on the server or client side. If we start going down this slope, we will never finish with "what if" scenarios, adding complexity to the code.

For these reasons, this max should be removed, eventually replaced by a log.Warning that an unexpected negative time has been detected. This log line should be added at the first moment where we can detect the problem, when we calculate sleepTime, before adding any jitter.

CHANGELOG.md
```diff
@@ -5,8 +5,20 @@ All notable changes to this project will be documented in this file.
...
+
+### Fixed
+
+- Fixes a bug with negative times that could occur due to the small difference between local (client) time and Github server time. Sleep time is now calculated using the date header (time as seen by the Github server). Additionally, negative sleep times are prevented by setting a minimum sleep time equal to the jitter time.
```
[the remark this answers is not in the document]
aliculPix4D: Changed as suggested.

cogito/ghcommitsink.go
```diff
@@ -38,6 +39,8 @@ func (sink GitHubCommitStatusSink) Send() error {
 		MaxRetries:   maxRetries,
 		WaitTime:     waitTime,
 		MaxSleepTime: maxSleepTime,
+		// adds some randomness to sleep time to prevent creating a Thundering Herd
```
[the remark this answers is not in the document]
aliculPix4D: Removed as suggested.

github/commitstatus.go

github/commitstatus.go
aliculPix4D: Fixed as suggested in the comment below.

github/commitstatus.go

github/commitstatus.go
aliculPix4D: Removed as suggested.

github/commitstatus.go
aliculPix4D: Still to do: 
- [ ] add warn log line if sleep time is negative

github/commitstatus.go
aliculPix4D: I added the link as comment.

aliculPix4D:
@marco-m-pix4d I addressed all your comments. Can you have a look again? 
I am espacially curious to hear your opinion on: https://github.com/Pix4D/cogito/pull/122/commits/26d0296e03f5c1f5c5af75393a7d662bc55c3dff

this is needed only to address your comment about negative sleep time:
>  log.Warning that an unexpected negative time has been detected

github/commitstatus.go
```diff
@@ -111,18 +113,23 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
...
 		if err != nil {
 			return err
 		}
-		timeToSleep = checkForRetry(response, s.target.WaitTime, s.target.MaxSleepTime)
-		if timeToSleep == 0 {
-			break
+		timeToSleep = checkForRetry(response, s.target.WaitTime, s.target.MaxSleepTime, s.target.Jitter)
+		switch {
+		case timeToSleep == 0:
+			break retryLoop
```
aliculPix4D, line 127: @marco-m-pix4d I didn't find a better way to break a loop. Is this in go language spirit?

github/commitstatus.go
```diff
@@ -111,18 +113,23 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
...
+		case timeToSleep < 0:
+			s.log.Warn("Sleep time should not have a negative value", "time", timeToSleep)
```
aliculPix4D, line 129: I was also thinking if we should just return an error here

github/commitstatus.go
marco-m-pix4d: if you use an `if` as opposed to a `switch`, you can use the `break` as before without the label at line 116, although the code as-is is correct.

But, I have the feeling we can simplify all this. Let me think a moment.

github/commitstatus.go
```diff
@@ -175,9 +183,16 @@ func httpRequestDo(req *http.Request) (httpResponse, error) {
...
-	response.rateLimitRemaining, _ = strconv.Atoi(resp.Header.Get("x-ratelimit-remaining"))
-	rateLimitReset, _ := strconv.Atoi(resp.Header.Get("x-ratelimit-reset"))
+	response.rateLimitRemaining, _ = strconv.Atoi(resp.Header.Get("X-RateLimit-Remaining"))
+	rateLimitReset, _ := strconv.Atoi(resp.Header.Get("X-RateLimit-Reset"))
 	response.rateLimitReset = time.Unix(int64(rateLimitReset), 0)
+
+	// time.RFC1123 is the format of the HTTP Date header
+	// example: Date:[ Mon, 02 Jan 2006 15:04:05 MST ]
+	// https://datatracker.ietf.org/doc/html/rfc2616#section-14.18
+	date, _ := time.Parse(time.RFC1123, resp.Header.Get("Date"))
```
aliculPix4D, line 193: same comment should apply here also: https://github.com/Pix4D/fly_helper/pull/158/files#r1201798389

> [marco-m-pix4d](https://github.com/marco-m-pix4d)(Marco Molteni) [2 minutes ago](https://github.com/Pix4D/fly_helper/pull/158/files#r1201798389)
In Go, never skip handling errors, unless there is a very specific reason.
This is equivalent to intercepting all exceptions in Python and dropping them.

github/commitstatus.go
```diff
@@ -175,9 +183,16 @@ func httpRequestDo(req *http.Request) (httpResponse, error) {
 	response.oauthInfo = fmt.Sprintf("X-Accepted-Oauth-Scopes: %v, X-Oauth-Scopes: %v",
 		resp.Header.Get("X-Accepted-Oauth-Scopes"), resp.Header.Get("X-Oauth-Scopes"))
 
-	response.rateLimitRemaining, _ = strconv.Atoi(resp.Header.Get("x-ratelimit-remaining"))
-	rateLimitReset, _ := strconv.Atoi(resp.Header.Get("x-ratelimit-reset"))
+	response.rateLimitRemaining, _ = strconv.Atoi(resp.Header.Get("X-RateLimit-Remaining"))
+	rateLimitReset, _ := strconv.Atoi(resp.Header.Get("X-RateLimit-Reset"))
```
aliculPix4D, line 187: same comment should apply here also: https://github.com/Pix4D/fly_helper/pull/158/files#r1201798389

> [marco-m-pix4d](https://github.com/marco-m-pix4d)(Marco Molteni) [2 minutes ago](https://github.com/Pix4D/fly_helper/pull/158/files#r1201798389)
In Go, never skip handling errors, unless there is a very specific reason.
This is equivalent to intercepting all exceptions in Python and dropping them.

aliculPix4D: This is not an excuse in any case, but I "learned" the best practice from:
https://github.com/google/go-github/blob/f96b64ece988554ba6444cb305e6db1132a61454/github/github.go#L674

marco-m-pix4d: From the documentation:

> The errors that ParseInt returns have concrete type *NumError and include err.Num = s. If s is empty or contains invalid digits, err.Err = ErrSyntax and the returned value is 0; if the value corresponding to s cannot be represented by a signed integer of the given size, err.Err = ErrRange and the returned value is the maximum magnitude integer of the appropriate bitSize and sign. 

So in case of error, the outcome of strconv.Atoi is `0`. They might have decided that `0` is OK, or they might have made a wrong assumption.

github/commitstatus.go
marco-m-pix4d:
```go
s.log.Warn("unexpected: negative sleep time", "time", timeToSleep)
```

marco-m-pix4d (approved):

aliculPix4D:
@Pix4D/integration request for a re-review. Thanks

github/commitstatus.go
aliculPix4D: All errors are now properly handled.

Esysc (approved):

github/commitstatus.go
```diff
@@ -175,9 +181,29 @@ func httpRequestDo(req *http.Request) (httpResponse, error) {
 	response.oauthInfo = fmt.Sprintf("X-Accepted-Oauth-Scopes: %v, X-Oauth-Scopes: %v",
 		resp.Header.Get("X-Accepted-Oauth-Scopes"), resp.Header.Get("X-Oauth-Scopes"))
 
-	response.rateLimitRemaining, _ = strconv.Atoi(resp.Header.Get("x-ratelimit-remaining"))
-	rateLimitReset, _ := strconv.Atoi(resp.Header.Get("x-ratelimit-reset"))
-	response.rateLimitReset = time.Unix(int64(rateLimitReset), 0)
+	if limit := resp.Header.Get("X-RateLimit-Remaining"); limit != "" {
```
aliculPix4D, line 184: I took some inspiration from: https://github.com/google/go-github/blob/9e4ce880cd4db6c8f8459806a8c1276a14db3983/github/github.go#L665

It seems that its perfectly valid and possible that these two headers (`X-RateLimit-Remaining`, `X-RateLimit-Reset`) are not reported. But if they are reported and we fail to convert their value to integer we must report an error.
In the `go-github` pkg they are happy enough with return value of 0 and ignore the error, I prefer to be explicit about it.

github/commitstatus.go
```diff
@@ -175,9 +181,29 @@ func httpRequestDo(req *http.Request) (httpResponse, error) {
...
+		v, err := strconv.Atoi(limit)
```
marco-m-pix4d, line 185: `v` is really too short also if this is Go, `val` is OK.
I find the "failed to convert" confusing: as a user (so not seeing the code), what does it mean?
It is enough to say: 
```go
fmt.Errorf("X-RateLimit-Remaining: %s", err)
```
which would print:
```
X-RateLimit-Remaining: strconv.Atoi: parsing "foo": invalid syntax
```
As proof, see https://go.dev/play/p/6K64il-jIE8

marco-m-pix4d: Consider also that strconv.Atoi is documented to return 0 in case of error, so we could simply write
```go
if limit := resp.Header.Get("X-RateLimit-Remaining"); limit != "" {
		response.rateLimitRemaining, err = strconv.Atoi(limit)
		if err != nil {
			return httpResponse{}, fmt.Errorf("X-RateLimit-Remaining: %s", err)
		}
}
```
or do as 1 line:
```
    // strconv.Atoi returns 0 in case of error.
	response.rateLimitRemaining, _ = strconv.Atoi(resp.Header.Get("X-RateLimit-Remaining"))
```

in the 3 cases, I think the result will be the same, AS LONG as this function is called on `response` only once. If it is called multiple times, then we must pick the correct behavior.

I don't have a strong opinion, as long as the code is consistent.

aliculPix4D: > Consider also that strconv.Atoi is documented to return 0 in case of error

but not always :)

```
val, err := strconv.Atoi("9223372036854775808")
err = fmt.Errorf("X-RateLimit-Remaining: %s", err)
fmt.Println(val, err)
```
results in:
`9223372036854775807 X-RateLimit-Remaining: strconv.Atoi: parsing "9223372036854775808": value out of range`

odormond (approved):

github/commitstatus.go
marco-m-pix4d: The example I gave was in case the header key was missing, which would result in a 0.

This code:
```go
if limit := resp.Header.Get("X-RateLimit-Remaining"); limit != "" {
    response.rateLimitRemaining, err = strconv.Atoi(limit)
    if err != nil {
        return httpResponse{}, fmt.Errorf("X-RateLimit-Remaining: %s", err)
    }
}
```
Can always be replaced by this:
```go
response.rateLimitRemaining, err = strconv.Atoi(limit)
if err != nil {
    return httpResponse{}, fmt.Errorf("X-RateLimit-Remaining: %s", err)
}
```

Regarding the out of range behavior, clamping to the biggest value encodable in the underlying int, seems quite reasonable to me, and is documented as such.

In any case, we need to keep the focus on the intent and not get lost on hypothetical corner cases.

What is the intent and the effect of returning in case of error?
Returning `httpResponse{}` will give the zero value to field `rateLimitRemaining`, which for ints is 0.
Is this what is wanted in case of error? Or is this expected not to be used because the caller is expected to handle the error and not look at `httpResponse` ?

PR #126, opened 2023-07-25
Update a bit everything

marco-m-pix4d:
In preparation of subsequent PRs. I also refreshed the GH token secret in CI, which had expired.

-    ci: update github actions
-    ci: update Task
-    test: replace tparse with gotestsum
-    go: update test dependencies
-    go: update dependencies
-    go: update go.mod to go 1.20, to match the Dockerfile and CI


iAmoric (approved):

SpectreVert (approved):

Taskfile.yml
```diff
@@ -27,16 +27,16 @@ vars:
...
+  GOTESTSUM_VERSION: v1.10.1
 
 tasks:
 
   install:deps:
     desc: Install tool dependencies.
     cmds:
       - go install github.com/golangci/golangci-lint/cmd/golangci-lint@{{.GOLANGCI_VERSION}}
-      - go install github.com/mfridman/tparse@{{.TPARSE_VERSION}}
+      - go install gotest.tools/gotestsum@{{.GOTESTSUM_VERSION}}
```
aliculPix4D, line 39: Can you write a few words on the reasoning here?

aliculPix4D (approved):

Taskfile.yml
marco-m-pix4d: The commit message for #79b9ce52c36 has an explanation. One problem is that tparse is used as a pipe, so if the previous stage (go test) fails, sometimes tparse cannot detect it (same problem with shell and -o pipefail). On the other hand gotestsum is used as a wrapper, so it can detect any failure of the child.

odormond (approved):

Issue #123, opened 2023-07-24
When retrying an HTTP request to GitHub, Cogito does not print why


marco-m-pix4d:
With the default log level, in case of retry, Cogito prints:

```
[INFO]  cogito.put.ghCommitStatus: GitHub HTTP request: attempt=1 max=3
[INFO]  cogito.put.ghCommitStatus: Sleeping for: time=4m47s
```

At line 2, the user can infer that the attempt at line 1 failed, but the reason (the HTTP status code) is not printed.
Instead, we should print the reason, because it would make easier to troubleshoot if needed.

PR #127, opened 2023-07-25
GitHub: log retry reason

marco-m-pix4d:
Closes: #123, PCI-3301

### DIFF
--- a/github/commitstatus_private_test.go
+++ b/github/commitstatus_private_test.go
@@ -16,13 +16,15 @@ func TestCheckForRetrySuccess(t *testing.T) {
 		maxSleepTime time.Duration
 		jitter       time.Duration
 		wantSleep    time.Duration
+		wantReason   string
 	}
 
 	run := func(t *testing.T, tc testCase) {
-		sleep, err := checkForRetry(tc.res, tc.waitTime, tc.maxSleepTime, tc.jitter)
+		sleep, reason, err := checkForRetry(tc.res, tc.waitTime, tc.maxSleepTime, tc.jitter)
 
 		assert.NilError(t, err)
 		assert.Equal(t, sleep, tc.wantSleep)
+		assert.Equal(t, reason, tc.wantReason)
 	}
 
 	testCases := []testCase{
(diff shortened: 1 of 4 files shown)


github/commitstatus_private_test.go
```diff
@@ -0,0 +1,80 @@
...
+			wantReason:   "rate limited",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) { run(t, tc) })
+	}
+}
+
+// BUG: sleeptime + jitter might cause a failure; test sleepTime > maxSleepTime should
```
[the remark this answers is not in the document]
marco-m-pix4d: That is something I noticed while writing the direct tests for checkForRetry. I wrote the comment as a reminder to write the corresponding test. Since it is only a comment with no tests it can be wrong. I would like not to discuss it here because it will derail this PR. We can discuss this in the PR that adds the test.

> The minimum value of jitter should always be >0

Should but is not.

github/commitstatus_private_test.go
```diff
@@ -0,0 +1,80 @@
...
+			maxSleepTime: 10 * time.Second,
+			wantSleep:    0 * time.Second,
+		},
+		{
+			name: "rate limited, would sleep a bit, adding also the jitter",
+			res: httpResponse{
+				statusCode: http.StatusForbidden,
+				date:       time.Date(2001, time.April, 30, 13, 0, 0, 0, time.UTC),
+				// 1h after
+				rateLimitReset: time.Date(2001, time.April, 30, 14, 0, 0, 0, time.UTC),
```
aliculPix4D, line 65: this works but it doesn't really make sense... Rate limit is reset each hour, so I would keep this value to <1h.
and also set `maxSleepTime` to a value smaller than 1h...

The code supports any values, that is the reason your tests passes, but this is how the server (Github) works so I would like to be in line with reality. It will be easier to understand later no?

aliculPix4D (approved):

github/commitstatus_private_test.go
marco-m-pix4d: The reason I wrote it in this way is exactly to avoid letting the reader thinking that the code knows anything about the GitHub time constants.

I understand it can be confusing; let me think about it.

marco-m-pix4d: Updated.

aliculPix4D: Thanks.

SpectreVert (approved):

Issue #125, opened 2023-07-24
report commit author and message


marco-m-pix4d:
Currently we send to the chat sink:

```
2023-07-24 09:04:37 UTC
pipeline foo-bar
job oo-bar-local-tests/1301
state 🔴 failure
commit 123456 (repo: example/repo)
```

It would be nice to send also the commit author and message (no matter the state):

```
2023-07-24 09:04:37 UTC
pipeline foo-bar
job oo-bar-local-tests/1301
state 🔴 failure
commit 123456 (repo: example/repo)
```

Same for the logging.

What we have:

```
[INFO]  cogito: This is the Cogito GitHub status resource. Tag: 0.10.1, commit: 05c92db, build date: 2023-05-31
[INFO]  cogito.put.ghCommitStatus: GitHub HTTP request: attempt=1 max=3
[INFO]  cogito.put.ghCommitStatus: commit status posted successfully: state=pending git-ref=123456
```

What we could have

```
[INFO]  cogito: This is the Cogito GitHub status resource. Tag: 0.10.1, commit: 05c92db, build date: 2023-05-31
[INFO]  cogito.put.ghCommitStatus: Processing commit git-ref=123456 author=johnny.stecchino comment="foo: add bar"
[INFO]  cogito.put.ghCommitStatus: GitHub HTTP request: attempt=1 max=3
[INFO]  cogito.put.ghCommitStatus: commit status posted successfully: state=pending git-ref=123456
```

PR #128, opened 2023-07-25
fix: internal error: unexpected: negative sleep time: 0s

### LINKED ISSUE #124: cogito: error: put: internal error: unexpected: negative sleep time: 0s
Just observed in our deployment:

```
[INFO]  cogito: This is the Cogito GitHub status resource. Tag: 0.10.1, commit: 05c92db, build date: 2023-05-31
[INFO]  cogito.put.ghCommitStatus: Sleeping for: time=4m47s
cogito: error: put: internal error: unexpected: negative sleep time: 0s
```

See also #123.

marco-m-pix4d:
Closes #124 and PCI-3302.

Best reviewed commit-per-commit and looking at the commit comments, otherwise it will be impossible to understand the details, and we do need to understand them in this case. Each commit tells the story and carefully builds on the precedent, keeping code coverage 100% (this doesn't ensure anything but is better than not doing it).

Last commit is perhaps the most unexpected, were we "define the errror out of existence", in the spirit of the book "A Philosophy of Software Design" by John Ousterhout (Talk: https://youtu.be/bmSAYlu0NcY?t=1313).

I have a recollection that the reason for the existence of the error is a comment I made on a previous PR by Aleksandar, so I am undoing what I proposed, because I learned from experience.

I decided not to touch the jitter initialization on purpose, because I think that the code in this PR is more robust and still as simple as it can. I am open to discussion.

### DIFF
--- a/github/commitstatus.go
+++ b/github/commitstatus.go
@@ -278,14 +278,19 @@ func checkForRetry(res httpResponse, waitTime, maxSleepTime, jitter time.Duratio
 		// Calculate the sleep time based solely on the server clock. This is unaffected
 		// by the inevitable clock drift between server and client.
 		sleepTime := res.rateLimitReset.Sub(res.date)
+		// Be robust to possible races in the GitHub backend. This avoids failing too early.
+		if sleepTime < 0 {
+			sleepTime = 0
+		}
 		// Be a good netizen by adding some jitter to the time we sleep.
 		sleepTime += jitter
+
 		switch {
 		case sleepTime == 0:
 			return 0, "", nil
 		case sleepTime > maxSleepTime:
 			return 0, "", nil
-		case sleepTime > 0 && sleepTime < maxSleepTime:
+		case sleepTime < maxSleepTime:
 			return sleepTime, "rate limited", nil
 		default:
 			return 0, "", fmt.Errorf("unexpected: negative sleep time: %s", sleepTime)
--- a/github/commitstatus.go
+++ b/github/commitstatus.go
@@ -286,7 +286,7 @@ func checkForRetry(res httpResponse, waitTime, maxSleepTime, jitter time.Duratio
 
 		switch {
 		case sleepTime == 0:
-			return true, 0, "rate limited, server-side inconsistency, should retry"
+			return true, 0, "rate limited, should retry"
 		case sleepTime > maxSleepTime:
 			return false, 0, "rate limited, sleepTime > maxSleepTime, should not retry"
 		case sleepTime < maxSleepTime:
--- a/github/commitstatus.go
+++ b/github/commitstatus.go
@@ -121,9 +121,9 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 		if err != nil {
 			return err
 		}
-		timeToSleep, reason := checkForRetry(response, s.target.WaitTime,
+		retry, timeToSleep, reason := checkForRetry(response, s.target.WaitTime,
 			s.target.MaxSleepTime, s.target.Jitter)
-		if timeToSleep == 0 {
+		if !retry {
 			break
 		}
 		s.log.Info("Sleeping for", "duration", timeToSleep, "reason", reason)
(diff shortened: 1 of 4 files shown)


github/commitstatus.go
```diff
@@ -262,35 +259,46 @@ func min(a, b int) int {
...
+	if res.statusCode == http.StatusForbidden && res.rateLimitRemaining == 0 {
 		// Calculate the sleep time based solely on the server clock. This is unaffected
 		// by the inevitable clock drift between server and client.
 		sleepTime := res.rateLimitReset.Sub(res.date)
+		// Be robust to possible races in the GitHub backend. This avoids failing too early.
+		if sleepTime < 0 {
+			sleepTime = 0
+		}
 		// Be a good netizen by adding some jitter to the time we sleep.
 		sleepTime += jitter
```
aliculPix4D, line 283: I will check again tomorrow but I think your PR still doesn't fix the situation that I think it happend (which is actually the simplest possible case)

- first request failed. Sleep time was around 4min30sec
- so we sleep this time
- second request is done. lets assume `res.rateLimitReset.Sub(res.date)` returned 0.
- by pure luck jitter rand func also returned 0
- even if we had your current code, third attempt would never happen because we brake out of the loop

odormond (approved):

github/commitstatus.go
odormond: You might not be talking to the same server on both attempts. They might have clocks off by just the right amount such that the second requests appears to arrive milliseconds before the reset.

But your point is still valid so probably making sure jitter is at least 1 would be an easy fix.

github/commitstatus.go
```diff
@@ -262,35 +259,46 @@ func min(a, b int) int {
...
+
 		switch {
+		case sleepTime == 0:
+			return 0, ""
```
aliculPix4D, line 287: I would add a msg here because this will terminate the loop without succesfull request, so the situation is quite different than:
```
case sleepTime > maxSleepTime:
   return 0, ""
```

github/commitstatus.go
marco-m-pix4d: @aliculPix4D 
I don't understand this scenario. Now that we have the dedicated unit test, can you come up with a test that shows the bug? If not, I think it is better to have a face chat so that I can understand the details.

marco-m-pix4d: > so probably making sure jitter is at least 1 would be an easy fix.

If the scenario turns out to be possible, I agree that this for sure is an easy fix.

github/commitstatus.go
marco-m-pix4d: There is no message because the current client code would not print such message.

> because this will terminate the loop without successful request

Ah. Let me check.

github/commitstatus.go
aliculPix4D: I would suggest you also change the jitter time to:
`5*time.Second + time.Duration(rand.Intn(30)) * time.Second`
https://github.com/Pix4D/cogito/blob/389e9a79e6622c6edc9fcdeb6119122a33b4db97/cogito/ghcommitsink.go#L43

github/commitstatus.go
aliculPix4D: Ok. I see why there is no msg.

But as we discussed in person, if the request failed, we would break the loop and the next attempt would not happen.
and in `checkStatus` we would inform the user we failed due to being rate limited and you had to wait just 0 seconds.

github/commitstatus.go
marco-m-pix4d: As discussed face-to-face, I did not change the jitter on purpose. The code must be correct also if jitter equals 0.

github/commitstatus.go
marco-m-pix4d: Resolving conversation as the code changed a lot, and the changes address also this.

marco-m-pix4d:
@Pix4D/integration: ready for re-review.
Thanks for @aliculPix4D for spotting the suble new bug of the subtle existing bug.
Please follow the new commits with their comments, they explain what is happening and step by step also simplify the logic.

aliculPix4D (approved):
LGTM.

You were correct, this is much better and way simpler. Thanks

Don't forget to close:
https://github.com/Pix4D/cogito/issues/124
https://github.com/Pix4D/cogito/issues/125

one is already solved, other is solved when you merge this PR.

marco-m-pix4d:
> Don't forget to close:
#124
#125

To double-check nothing is missing:

since #124 is mentioned in this PR (or in a commit comment), it will be closed automatically:
pr

while #125 is unrelated.

In my understanding, there is nothing to do. Am I missing something?

aliculPix4D:
> In my understanding, there is nothing to do. Am I missing something?

no, it was my mistake: I was thinking of #/123 and for the other #124, I didn't know, it will be closed automatically:

PR #129, opened 2023-07-26
Refactor github/commitstatus tests

marco-m-pix4d:
Since to address the bug found by @aliculPix4D in #128 I want to add a repro test, I am doing this pure refactoring PR first, to reduce spurious diffs.

- test: github/commitstatus: refactor: use gotest.tools. We are already importing the gotest.tools package in cogito, but this test file was not yet using it.
- test: github/commitstatus: refactor: use same testing pattern. I found the testing pattern for table-driven tests suggested by the gotest.tools author, https://github.com/gotestyourself/gotest.tools/wiki/Go-Testing-Patterns#table-driven-tests, to be way more readable than that standard approach.

### DIFF
--- a/github/commitstatus_test.go
+++ b/github/commitstatus_test.go
@@ -24,6 +24,11 @@ type mockedResponse struct {
 }
 
 func TestGitHubStatusSuccessMockAPI(t *testing.T) {
+	type testCase struct {
+		name     string
+		response mockedResponse
+	}
+
 	cfg := testhelp.FakeTestCfg
 	context := "cogito/test"
 	targetURL := "https://cogito.invalid/builds/job/42"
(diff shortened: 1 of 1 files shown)


odormond (approved):

github/commitstatus_test.go
```diff
@@ -192,23 +192,16 @@ OAuth: X-Accepted-Oauth-Scopes: , X-Oauth-Scopes: `,
...
 			var ghError *github.StatusError
 			if !errors.As(err, &ghError) {
 				t.Fatalf("\nhave: %s\nwant: type github.StatusError", err)
 			}
 			wantStatus := tc.response.statuses[len(tc.response.statuses)-1]
-			if have, want := ghError.StatusCode, wantStatus; have != want {
-				t.Fatalf("status code: have: %d; want: %d", have, want)
-			}
-
-			if diff := cmp.Diff(wantErr, err.Error()); diff != "" {
```
aliculPix4D, line 209: is this a mistake? why you don't want to check if the errors match? if you really want to remove that assert you can also cleanup all key values of `wantErr` in test cases.

github/commitstatus_test.go
```diff
@@ -315,20 +307,12 @@ OAuth: X-Accepted-Oauth-Scopes: , X-Oauth-Scopes: repo:status`,
...
+			assert.Error(t, err, tc.wantErr)
 			var ghError *github.StatusError
 			if !errors.As(err, &ghError) {
 				t.Fatalf("\nhave: %s\nwant: type github.StatusError", err)
 			}
-			if have, want := ghError.StatusCode, tc.wantStatus; have != want {
-				t.Fatalf("status code: have: %d; want: %d", have, want)
-			}
-
-			if diff := cmp.Diff(tc.wantErr, err.Error()); diff != "" {
```
aliculPix4D, line 329: same question here...

aliculPix4D:
So I think that you either did a mistake in the first commit or if that was intended (personally I don't think so) there is more cleanup to do.

github/commitstatus_test.go
marco-m-pix4d: Thanks for looking closely at the code :-)

In this case, it should be correct, because

```go
assert.Error(t, err, wantErr)
```

replaces 2 asserts:
```go
if err == nil {
				t.Fatalf("\nhave: <no error>\nwant: %s", wantErr)
}

if diff := cmp.Diff(wantErr, err.Error()); diff != "" {
				t.Fatalf("error: (+have -want):\n%s", diff)
}
```

aliculPix4D: Cool... Thanks

aliculPix4D (approved):

PR #133, opened 2023-08-08
fix #132: does not sleep when it should, approach 1

### LINKED ISSUE #132: retry GitHub HTTP request when rate limited: does not sleep when it should
We detect the situation correctly, we calculate the correct time to sleep, we log that we will sleep but actually we do NOT sleep, thus we consume the 3 attempts one after the other and finally fail:

```
[INFO]  cogito: This is the Cogito GitHub status resource. Tag: 0.10.2, commit: d8221ad, build date: 2023-07-28
[INFO]  cogito.put.gChat: not sending to chat: reason="feature not enabled"
[INFO]  cogito.put.ghCommitStatus: Sleeping for: duration=7m42s reason="rate limited, should retry"
[INFO]  cogito.put.ghCommitStatus: Sleeping for: duration=7m42s reason="rate limited, should retry"
[INFO]  cogito.put.ghCommitStatus: GitHub HTTP request: attempt=3 max=3
[INFO]  cogito.put.ghCommitStatus: Sleeping for: duration=7m42s reason="rate limited, should retry"

cogito: error: put: failed to add state "pending" for commit XXX: 403 Forbidden
Body: {"message":"API rate limit exceeded for user ID YYY.","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
Hint: Rate limited but the wait time to reset would be longer than 15m0s (MaxSleepTime)
```

This has been introduced by #128 

😿

marco-m-pix4d:
As mentioned, this is approach 1 at fixing #132. See also approach 2 at #134.

It works, but it makes the test to take longer (as they were taking before I added this bug) and is somehow flaky, since it is obliged to measure time, and time can skew based on CI load.

Best reviewed commit-per-commit, since commit 1 adds the repro and commit 2 adds the fix.

Fixes #132.

### DIFF
--- a/github/commitstatus.go
+++ b/github/commitstatus.go
@@ -112,10 +112,8 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 	req.Header.Set("Content-Type", "application/json")
 
 	var response httpResponse
-	timeToSleep := 0 * time.Second
 
 	for attempt := 1; attempt <= s.target.MaxRetries; attempt++ {
-		time.Sleep(timeToSleep)
 		s.log.Info("GitHub HTTP request", "attempt", attempt, "max", s.target.MaxRetries)
 		response, err = httpRequestDo(req)
 		if err != nil {
@@ -127,6 +125,7 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 			break
 		}
 		s.log.Info("Sleeping for", "duration", timeToSleep, "reason", reason)
+		time.Sleep(timeToSleep)
 	}
 
 	return s.checkStatus(response, state, sha, url)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v0.10.3] - 2023-08-XX
+
+### Fixed
+
+- Retry GitHub HTTP request when rate limited: the fix for [#124](https://github.com/Pix4D/cogito/issues/124) introduced a new bug. This fixes it. 
+
 ## [v0.10.2] - 2023-07-28
 
 ### Fixed
 
-- Retry GitHub HTTP request when rate limited: fix internal error: unexpected: negative sleep time: 0s
+- Retry GitHub HTTP request when rate limited: fix internal error: unexpected: negative sleep time: 0s [#124](https://github.com/Pix4D/cogito/issues/124).
 
 ### Changed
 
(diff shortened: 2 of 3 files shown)


github/commitstatus.go
```diff
@@ -112,10 +112,8 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 	req.Header.Set("Content-Type", "application/json")
 
 	var response httpResponse
-	timeToSleep := 0 * time.Second
 
 	for attempt := 1; attempt <= s.target.MaxRetries; attempt++ {
-		time.Sleep(timeToSleep)
```
aliculPix4D, line 118: When I did my previous implementation there were two hidden marvels in this for loop.

1. avoid shadowing `:=` 
2. put `time.Sleep(timeToSleep)` at the top and not at the bottom.

At the time, I couldn't make it to work when I added `time.Sleep(timeToSleep)` at the end of the iteration like you are doing now. Unfortunately, I forgot what was the reason and now looking at your code I can't find any reason why it wouldn't work as proposed in this PR. 

I will go back a bit to the previous version of the code, to try to identify the reason. I just wanted to mention here to be careful a bit.

marco-m-pix4d: I _think_ it was only to avoid a spurious sleep(0), although it would have been harmless.

marco-m-pix4d: See also especially approach 2, where the tests are actually testing everything and should make it clear that the code is now correct.

aliculPix4D: Ok. This can't occur with your code.

It was specific to that implementation at the time.
Here is the discussion: https://github.com/Pix4D/cogito/pull/118#discussion_r1173377510
and commit that moved it up: https://github.com/Pix4D/cogito/commit/c8f7ab2d1fbd2f8ea898ee590f2500d2b7d5fe71

aliculPix4D: actually, wait... it's still important. If max attempt is reached you will sleep for `waitTime` but there will not be next attempt. in example, for 500 status codes.

marco-m-pix4d: Ah. I think I understand what you mean. We are missing a test also for this case.

The more I touch this code, the more I think that an exponential back-off with a stop only on a time limit, no matter the number of attempts, would be simpler and so with less corner cases. We went with counting the attempts in an effort to "optimize" the number of requests towards GitHub since we knew when the reset would happen, but this caused the code to be fragile. Such optimization is not needed.

I will add the missing test and a fix for this behavior.

aliculPix4D: Note also, that my preferred approach is option 2. So I will continue with comments there, and just "ignore" this PR.

marco-m-pix4d: For sure. This approach is here to show how to evolve an approach to something better.

marco-m-pix4d:
Closed in favor of #134

PR #134, opened 2023-08-08
fix #132: does not sleep when it should, approach 2

### LINKED ISSUE #132: retry GitHub HTTP request when rate limited: does not sleep when it should

marco-m-pix4d:
As mentioned, this is approach 2 at fixing https://github.com/Pix4D/cogito/issues/132. See also approach 1 at #133.

Compared with #133, it keeps the tests running as fast as possible and does not introduce test flakiness since it does not measure time.

To do this, I added an injectable sleep function, which I then substituted with a spy in the tests.

Best reviewed commit-per-commit, since commit 1 adds the repro and commit 2 adds the fix.

Fixes https://github.com/Pix4D/cogito/issues/132.


github/commitstatus_test.go
```diff
@@ -32,6 +32,10 @@ func TestGitHubStatusSuccessMockAPI(t *testing.T) {
 	type testCase struct {
 		name     string
 		response []mockedResponse
+		// wantSleeps:
```
marco-m-pix4d, line 35: see the comments here @aliculPix4D

SpectreVert (approved):

github/commitstatus.go
```diff
@@ -127,6 +135,7 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
 			break
 		}
 		s.log.Info("Sleeping for", "duration", timeToSleep, "reason", reason)
+		s.sleepFn(timeToSleep)
```
aliculPix4D, line 138: See here: https://github.com/Pix4D/cogito/pull/133/files#r1286852612

I think this is still problematic.

github/commitstatus_test.go
```diff
@@ -113,6 +125,7 @@ func TestGitHubStatusSuccessMockAPI(t *testing.T) {
 					rateLimitReset:     now.Add(1 * time.Hour).Unix(),
 				},
 			},
+			wantSleeps: []time.Duration{0 * time.Second},
 		},
 		{
```
aliculPix4D, line 130: I remember having a test when max attempts were reached. Can we add such a test here if it was removed?

aliculPix4D (commented):
So I prefer if we continue with the approach shown in this PR. I am not approving yet only because there is still an issue with un-necessary sleep at max retries and a missing test for it. 
I am just puzzled why the test with max retries is missing, when I am 100% sure I had it at the time.

github/commitstatus_test.go
marco-m-pix4d: I have no recollection of having removed a test, although for sure I might have by mistake.

I will look into this and I also plan to make the tests more similar to prod (number of attempts = 3), because now they require to keep in mind that number of attempts = 2 to understand the failure logic. Now that the sleep is virtual, we can attempt = 3 as in prod.

aliculPix4D: So I think this is the point. You added tests only to: `TestGitHubStatusSuccessMockAPI`
and you can catch this error only in `TestGitHubStatusFailureMockAPI` if you add do the same changes in that test and "measure" the time.

marco-m-pix4d: Yes, I can now reproduce on that test indeed, thanks.

github/commitstatus.go
marco-m-pix4d: Thanks. Fixed.

github/commitstatus_test.go
marco-m-pix4d: Fixed.

marco-m-pix4d:
@Pix4D/integration: ready for re-review.
Given the refactoring, I suggest to review commit-per-commit; the commit message is there to help understanding the reason of each change.

Thanks @aliculPix4D for pointing out yet another corner case. It has been longer than expected, but I think now that we can trust the tests. I also think that the retry loop is now self-evident.

SpectreVert (approved):

github/commitstatus_test.go
```diff
@@ -78,22 +103,23 @@ func TestGitHubStatusSuccessMockAPI(t *testing.T) {
...
-					rateLimitReset: now.Add(1 * time.Second).Unix(),
+					rateLimitReset:     now.Add(42 * time.Second).Unix(),
 				},
 				{
 					status:             http.StatusCreated,
 					rateLimitRemaining: fullRateRemaining,
 					rateLimitReset:     now.Add(1 * time.Hour).Unix(),
 				},
 			},
+			wantSleeps: []time.Duration{42 * time.Second},
```
aliculPix4D, line 122: What do you think about adding 1 more test, with jitter this time? I think jitter is implicitly set to 0 currently.

github/commitstatus_test.go
```diff
@@ -55,17 +74,23 @@ func TestGitHubStatusSuccessMockAPI(t *testing.T) {
 		}
 		ts := httptest.NewServer(http.HandlerFunc(handler))
 		defer ts.Close()
-
 		target := github.Target{
```
aliculPix4D, line 77: jitter is implicitly set to 0 here. Do we want to be explicit about it?

github/commitstatus_test.go
```diff
@@ -162,15 +193,19 @@ func TestGitHubStatusFailureMockAPI(t *testing.T) {
...
 		ts := httptest.NewServer(http.HandlerFunc(handler))
 		defer ts.Close()
-
 		wantErr := fmt.Sprintf(tc.wantErr, ts.URL)
 		target := github.Target{
-			Server:       ts.URL,
-			MaxRetries:   2,
-			WaitTime:     time.Second,
-			MaxSleepTime: maxSleepTime,
+			Server:              ts.URL,
```
aliculPix4D, line 198: same here. Jitter is implicitly set to 0.

github/commitstatus_test.go
odormond: I agree with you. We don't actually test that line 309 `sleepTime += jitter` is there.

aliculPix4D (approved):
Pre-approving. I would just handle the case where `jitter>0` also.

github/commitstatus_test.go
marco-m-pix4d: Jitter is covered in the unit tests for checkForRetry, see for example https://github.com/Pix4D/cogito/blob/master/github/commitstatus_private_test.go#L66. As such, I think this is OK as-is, unless I am missing something?

github/commitstatus_test.go
marco-m-pix4d: See thread just above (https://github.com/Pix4D/cogito/pull/134#discussion_r1288255313)

github/commitstatus_test.go
aliculPix4D: > I think this is OK as-is, unless I am missing something.

I also think so from the testing side. From the clarity point of view, I am still wondering if its worth adding an explicit field for jitter when initializing `github.Target{}` to improve test clarity and avoid people asking questions like mine.

On the other hand, this is normal Go behavior; initialize to the 0 value of that type.

odormond: > Jitter is covered in the unit tests for `checkForRetry`, see for example https://github.com/Pix4D/cogito/blob/master/github/commitstatus_private_test.go#L66. As such, I think this is OK as-is, unless I am missing something?

All good then. I only checked the test files modified in the PR. Sorry.

marco-m-pix4d: > On the other hand, this is normal Go behavior; initialize to the 0 value of that type.

Exactly, this is why I do not touch these fields.

There are other cases where I do state the default, for example

```go
wantSleeps: nil,
```

is technically useless also. I make it explicit because I think we want to make it obvious that we expect 0 sleeps in that test case.

The jitter in my opinion is different: we simply do not care (in these tests).

At the end, it is subjective. I'd rather leave as-is for these reasons. If you are uneasy, I can make the jitter explicit too.

aliculPix4D: > At the end, it is subjective. I'd rather leave as-is for these reasons. If you are uneasy, I can make the jitter explicit too.

No it's fine as it is. I was also unsure if it's required or not when I asked the question, so I wanted to double-check and see your opinion on it.

PR #139, opened 2023-09-01
Support GitHub Enterprise (GHE) - attempt #2

wanderanimrod:
_pairing with @dev-jin_

# Why
This resource only supports setting status on repos hosted in public GitHub. The domain `github.com` and `api.github.com` was the only API endpoint to which this resource could post-commit statuses.

# What
1. Add the field `github_api_endpoint` to the resource's source config (optional)
2. Make `GitHubCommitStatusSink` and `ProdPutter` use the new source config field when present. If not, keep defaulting to the public GitHub API. Making the `github_api_endpoint` config optional keeps the resource simple for the majority of users (who I assume use public GH) and also ensures backward compatibility when those users upgrade to a version of this resource that supports GHE
3. Update the README to mention the new source config field `github_api_endpoint`.
4. Housekeeping
  a. Remove the ability to override the GH API with the `COGITO_GITHUB_API` environment variable. Change tests that used this to use the new `github_api_endpoint` source config field.
  b. Minor changes like fixing typos, removing unused vars and imports, etc. See inline comments for details.

# Tests
- I have tested this against GitHub Enterprise version `3.3.19 `with the `github_api_endpoint` source param set to my company's GitHub hostname and it works.


cogito/protocol.go
```diff
@@ -254,10 +257,18 @@ func (src *Source) Validate() error {
...
+	if src.GithubApiEndpoint != "" {
+		u, err := url.ParseRequestURI(src.GithubApiEndpoint)
+		if err != nil || u.Host == "" {
+			return fmt.Errorf("source: github_api_endpoint '%s' is an invalid api endpoint", src.GithubApiEndpoint)
+		}
+		if u.Scheme != "https" { //catch http prefix that falls through
+			return fmt.Errorf("source: github_api_endpoint %s must have a https prefix, not http", src.GithubApiEndpoint)
+		}
+	}
+
```
wanderanimrod, line 270: Just some validation to ensure users only pass valid URLs to the resource. If we don't catch this early, the errors thrown by the resource become cryptic, making it harder for users to figure out what they did wrong.

cogito/protocol_test.go
```diff
@@ -192,7 +240,8 @@ log_level:             debug
 context_prefix:        the-prefix
 chat_append_summary:   true
 chat_notify_on_states: [success failure]
-sinks: []`
+sinks: []
+github_api_endpoint: `
```
wanderanimrod, line 244: I wasn't too sure about what to do here. Most people won't configure the `github_api_endpoint` so the field will be printed as empty. Looks strange, but I needed it to make this test pass.

cogito/put.go
```diff
@@ -40,7 +38,7 @@ type Sinker interface {
 // Additionally, the script may emit metadata as a list of key-value pairs. This data is
 // intended for public consumption and will make it upstream, intended to be shown on the
 // build's page.
-func Put(log hclog.Logger, input []byte, out io.Writer, args []string, putter Putter) error {
```
wanderanimrod, line 43: This function parameter was not used. Related imports were also removed.

cogito/putter.go
```diff
@@ -262,8 +273,9 @@ func checkGitRepoDir(dir, owner, repo string) error {
 	if err != nil {
 		return fmt.Errorf(".git/config: remote: %w", err)
 	}
-	left := []string{"github.com", owner, repo}
```
wanderanimrod, line 265: Repositories hosted on GitHub enterprise have unpredictable "hostnames" in their git config. I couldn't think of a way to generically test what the value of the field should be, and I couldn't think about why we would need to test on this now that we support a broader set of GitHub hosts. So, I removed it the hostname validation part and kept the rest. 

Is this okay?

cmd/cogito/main.go
```diff
@@ -50,21 +50,14 @@ func mainErr(in io.Reader, out io.Writer, logOut io.Writer, args []string) error
 	})
 	log.Info(cogito.BuildInfo())
 
-	ghAPI := os.Getenv("COGITO_GITHUB_API")
-	if ghAPI != "" {
-		log.Info("endpoint override", "COGITO_GITHUB_API", ghAPI)
-	} else {
-		ghAPI = github.API
-	}
-
```
wanderanimrod, line 59: I removed the environment-based API endpoint override. There was only one usage of it (in`main_test.go`). I replaced that usage with the custom test URL in the `github_api_endpoint` source config. The test still works.

Let me know if you have other plans, and I will revert this change.

cmd/cogito/main_test.go
```diff
@@ -69,19 +69,19 @@ func TestRunPutSuccess(t *testing.T) {
...
+			GChatWebHook:      googleChatSpy.URL,
+			LogLevel:          "debug",
 		},
 		Params: cogito.PutParams{State: wantState},
 	}))
 	var out bytes.Buffer
 	var logOut bytes.Buffer
 	inputDir := testhelp.MakeGitRepoFromTestdata(t, "../../cogito/testdata/one-repo/a-repo",
 		testhelp.HttpsRemote("the-owner", "the-repo"), "dummySHA", wantGitRef)
-	t.Setenv("COGITO_GITHUB_API", gitHubSpy.URL)
```
wanderanimrod, line 84: Moved this into the source config.

marco-m-pix4d:
Hello @wanderanimrod happy to see your persistence! :-)
Please be advised that we are under load in my team so it might take a while for a review.

wanderanimrod:
> Hello @wanderanimrod happy to see your persistence! :-)
> Please be advised that we are under load in my team so it might take a while for a review.

Hello @marco-m-pix4d , any chance this will get looked at anytime soon?

README.md
```diff
@@ -198,6 +198,10 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
   The log level (one of `debug`, `info`, `warn`, `error`, `silent`).\
   Default: `info`.
 
+- `github_api_endpoint`:\
+  Override the default API endpoint. This allows you to post commit statuses to repos hosted by GitHub Enterprise (GHE) instances. When provided, this parameter must be a valid https url.
```
aliculPix4D, line 202: repos --> repositories

cogito/buildinfo.go
```diff
@@ -1,11 +1,11 @@
 package cogito
 
 // Baked in at build time with the linker. See the Taskfile and the Dockerfile.
-var buildinfo = "unknown"
+var buildInfo = "unknown"
```
aliculPix4D, line 4: this required a change in:
https://github.com/Pix4D/cogito/blob/9558fa3fe5454cc1e1ff44c144cef9786862485b/Taskfile.yml#L22
https://github.com/Pix4D/cogito/blob/9558fa3fe5454cc1e1ff44c144cef9786862485b/Dockerfile#L24

cmd/cogito/main.go
```diff
@@ -50,21 +50,14 @@ func mainErr(in io.Reader, out io.Writer, logOut io.Writer, args []string) error
...
 	switch cmd {
 	case "check":
 		return cogito.Check(log, input, out, args[1:])
 	case "in":
 		return cogito.Get(log, input, out, args[1:])
 	case "out":
-		putter := cogito.NewPutter(ghAPI, log)
-		return cogito.Put(log, input, out, args[1:], putter)
+		putter := cogito.NewPutter(github.API, log)
```
aliculPix4D, line 59: it feels strange that we need to pass github api endpoint already here for the `cogito.NewPutter`, given that one can only use Gchat sink for example, that doesn not require any knowleage of the Github side.

For me, this seems a remnant of the past... I wonder if we can improve here.

cmd/cogito/main.go
```diff
@@ -64,7 +65,7 @@ func mainErr(in io.Reader, out io.Writer, logOut io.Writer, args []string) error
 		return cogito.Get(log, input, out, args[1:])
 	case "out":
 		putter := cogito.NewPutter(ghAPI, log)
-		return cogito.Put(log, input, out, args[1:], putter)
```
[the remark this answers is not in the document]
aliculPix4D: Ok. I see... this was leftover from the invocation was consistent for all 3 commands. Now, you are correct we never invoke the logger sent as function parameter. Only the logger from putter interface is used.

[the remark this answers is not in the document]
aliculPix4D: TODO
- [ ] to check why this isnt catched by golangci-lint (https://github.com/Pix4D/cogito/blob/9558fa3fe5454cc1e1ff44c144cef9786862485b/Taskfile.yml#L44).

cogito/put_test.go
```diff
@@ -405,6 +436,25 @@ func TestPutterSinks(t *testing.T) {
 	assert.Assert(t, ok2)
 }
 
+func TestGitHubCommitStatusSinkApiEndpointOverrideFromSource(t *testing.T) {
```
aliculPix4D, line 439: these are two tests in one... should table driven tests be used here?

cogito/putter.go
aliculPix4D: Just a bit of background here first. This part has two purposes:
- ensure that the git resource directory given as input to cogito matches the one configured in cogito source configuration. We must ensure that the information parsed "locally" from the git directory matches the information configured in the source. That means all: endpoint, organization, and repository name
- give the best possible information to the end user. People usually copy-paste the Concourse YAML configuration from other existing pipelines. That means that they copy-paste the cogito source configuration, and while they change the majority of the fields sometimes they forget some or make a typo.
The error msg like should ensure that the mistake is easily and quickly identified:
```
The received git repository is incompatible with the Cogito configuration.
Git repository configuration (received as 'inputs:' in this PUT step):
 url: git@<input-hostname>:<input-org-name>/<input-repo-name>.git
 owner: <input-org-name>
 repo: <input-repo-name>
Cogito SOURCE configuration:
 owner: <source-org-name>
 repo: <source-repo-name>
```

So instead of removing one check for the endpoint, you now actually MUST perform this check because if one configures `github_api_endpoint` in source configuration, you would expect this to match to what you parse from your local `.git/config`. 

To summarize, one must validate that:
- <input-hostname> == <source-github_api_endpoint>
- <input-org-name> == <source-org-name>
- <source-repo-name> == <source-repo-name>

cogito/protocol_test.go
aliculPix4D: This wouldn't be an issue if you added the default endpoint during JSON parsing:
https://github.com/Pix4D/cogito/blob/6930bc4eb07148496d608fed809ff4762964eccf/cogito/protocol.go#L175
or to be consistent with other fields in `func (src *Source) Validate() error`
https://github.com/Pix4D/cogito/blob/6930bc4eb07148496d608fed809ff4762964eccf/cogito/protocol.go#L271
in if github endpoint is not set; set the default.

I think this is much better approach then sending the some API endpoint from the top (main during putter creation)

aliculPix4D (commented):
First, apologies for the long wait. Our team was swamped by urgent issues last month.

Overall very nice work. Thank you for the effort and your PR.

I think we can simplify this a bit further. Let's wait the comments from @marco-m-pix4d

README.md
wanderanimrod, line 202:
```suggestion
  Override the default API endpoint. This allows you to post commit statuses to repositories hosted by GitHub Enterprise (GHE) instances. When provided, this parameter must be a valid https url.
```

wanderanimrod (commented):
thanks for the feedback @aliculPix4D . I have addressed the simple issues. will work on the more complex changes (the unresolved threads) soon and notify you.

cogito/buildinfo.go
wanderanimrod: This was reverted in 558f12d778497501ba0122ea58436672971dd71c. It was gong to be a larger change than I expected, yet it has nothing to do with GHE.

marco-m-pix4d:
@aliculPix4D if you need help on how to proceed, please ping me on our chat.

aliculPix4D:
Closing this PR in favor of: https://github.com/Pix4D/cogito/pull/143

Note that the first commit is the squash of all commits present in this PR and it gives the credit to both authors from this PR.

aliculPix4D:
@wanderanimrod just a bit more context. Basically, the only reason I closed this PR is because it was created from your fork, and it was simple for me to commit and test the changes in this repo. So creating a new PR with a branch in this repository simplified running the tests and especially creating Concourse pipelines for our acceptance tests.

wanderanimrod:
Thanks for the explanation @aliculPix4D . Makes sense. Looking forward to #143 getting merged. This has been on our backlog for a long time. Glad to see it coming to an end.

PR #140, opened 2023-09-15
New generic customizable retry mechanism, independent from GitHub

marco-m-pix4d:
Best reviewed commit per commit:
- first we introduce the new retry package
- then we integrate it in cogito

Note that:

1. In the name of resilience, I changed the behavior in case of zero or negative sleep time: we do not try any more to be smart, we fall-back to the default back-off sequence. In particular, have a close look at the tests. I arrived at this approach after many attempts, but for sure I might have missed something, so we can change if needed.
2. To keep things simple and incremental, the new retry mechanism does not support jitter. I think this can be safely added back later.

Part of PCI-3309.


github/commitstatus_private_test.go
```diff
@@ -1,114 +1,106 @@
 package github
 
-import (
```
marco-m-pix4d, line 3: ah, I had forgotten this file. I think we can just delete it, but I will double-check later.

cogito/ghcommitsink.go
```diff
@@ -1,26 +1,27 @@
...
-	// limited from GitHub.
-	maxSleepRateLimited = 15 * time.Minute
+	// retryFirstDelay is duration of the first backoff.
+	retryFirstDelay = 2 * time.Second
 
-	// waitTransient is the wait time before the next attempt when encountering a
-	// transient error from GitHub.
-	waitTransient = 5 * time.Second
+	// retryBackoffLimit is the upper bound duration of a backoff.
+	// That is, with an exponetial backoff and a retryFirstDelay = 2s, the sequence will be:
```
aliculPix4D, line 22: typo: exponetial

github/retry.go
```diff
@@ -0,0 +1,82 @@
...
+func TransientError(statusCode int) bool {
+	switch statusCode {
+	case
+		http.StatusInternalServerError, // 500
+		http.StatusBadGateway,          // 502
+		http.StatusServiceUnavailable,  // 503
+		http.StatusGatewayTimeout:      // 504
+		return true
+	default:
+		return false
```
odormond, line 80: So this means on a connection timeout we would get a HardFail and not retry the request, right? If so, that's a pity because that's the only case when it's always safe to retry without even prior knowledge of the API.

marco-m-pix4d: Ah, I did not consider that. I believe that this was also the case before.
Anyway, here we are testing at the HTTP level. To do what you are proposing, we would have to test also at the L3/TCP, which should be possible with this API, it still looks into an error ...

I will create a ticket

marco-m-pix4d: https://pix4dbug.atlassian.net/browse/PCI-3387

odormond (approved):

cogito/ghcommitsink.go
```diff
@@ -1,26 +1,27 @@
...
 
 import (
-	"math/rand"
+	"log/slog"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
 
 	"github.com/Pix4D/cogito/github"
+	"github.com/Pix4D/cogito/internal/hclogslog"
```
aliculPix4D, line 10: why are we mixing `log/slog` and `github.com/Pix4D/cogito/internal/hclogslog`?

internal/hclogslog/adapter.go
```diff
@@ -0,0 +1,159 @@
+// Taken from https://github.com/evanphx/go-hclog-slog and modified to use
```
aliculPix4D, line 1: I must say; I am not a fan of this...

aliculPix4D: Also, why internal, why not vendored? 

Is the modification done by you or the author?

retry/retry.go
```diff
@@ -0,0 +1,97 @@
...
+	if rtr.FirstDelay <= 0 {
+		return errors.New("FirstDelay must be positive")
+	}
+	if rtr.BackoffLimit <= 0 {
+		return errors.New("BackoffLimit must be positive")
+	}
+	if rtr.SleepFn == nil {
+		rtr.SleepFn = time.Sleep
+	}
+	rtr.Log = rtr.Log.With("system", "retry") // FIXME maybe better constructor???
```
aliculPix4D, line 69: Can you explain the comment a bit? > FIXME maybe better constructor???
What needs to be fixed and why? in other words what you don't like here?

retry/retry.go
```diff
@@ -0,0 +1,97 @@
...
+
+	delay := rtr.FirstDelay
+	totalDelay := 0 * time.Second
+
+	for attempt := 1; ; attempt++ {
+		err := workFn()
+		switch classifierFn(err) {
+		case Success:
+			rtr.Log.Info("success", "attempt", attempt, "totalDelay", totalDelay)
+			return err
```
aliculPix4D, line 79: This is confusing at first. So err is nil in this case and we just return nil?

retry/retry_example_test.go
```diff
@@ -0,0 +1,135 @@
...
+		Log: slog.New(slog.NewTextHandler(os.Stdout,
+			&slog.HandlerOptions{ReplaceAttr: removeTime})),
+	}
+
+	workFn := func() error {
+		// Do work...
+		return nil
+	}
+	classifierFn := func(err error) retry.Action {
+		if err != nil {
```
aliculPix4D, line 32: is this needed in this example? How can you enter here in this specific example?

retry/retry_test.go
```diff
@@ -0,0 +1,147 @@
...
+import (
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/go-quicktest/qt"
```
aliculPix4D, line 12: why? experiment? I am a bit confused here, don't we change this test package two often now...

What about `gotest.tools/v3/assert`? If we introduce the new one, shouldn't we make it consistent everywhere....

github/commitstatus_private_test.go
```diff
@@ -1,114 +1,106 @@
...
-			wantSleep:  1 * time.Second,
-			wantReason: "rate limited, should retry",
-		},
-	}
-
-	for _, tc := range testCases {
-		t.Run(tc.name, func(t *testing.T) { run(t, tc) })
-	}
-}
+// func TestCheckForRetry(t *testing.T) {
```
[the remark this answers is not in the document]
aliculPix4D: Ah... I see: https://github.com/Pix4D/cogito/pull/140/files#r1327110237

SpectreVert (approved):

github/commitstatus.go
```diff
@@ -129,197 +120,71 @@ func (s CommitStatus) Add(sha, state, targetURL, description string) error {
...
-func (s CommitStatus) checkStatus(resp httpResponse, state, sha, url string) error {
-	var hint string
-
-	switch resp.statusCode {
-	case http.StatusCreated:
-		// Happy path
-		return nil
-	case http.StatusNotFound:
-		hint = fmt.Sprintf(`one of the following happened:
+// TODO: can we merge (at least partially) this function in GitHubError.Error ?
```
aliculPix4D, line 152: I see that you have similar questions I had...

I would very much like to get rid of this function/method or rewrite it nicer, but at the time I could not find any reasonable solution for it...

internal/hclogslog/adapter.go
marco-m-pix4d: Actually it is vendored, since it is here in the code and not pulled in as a dependency. It cannot be put in the vendor directory because it is modified.
It is modified by me; if it were modified by the author, I would pull the dependency instead :-)

If you explain why you are not a fan, I can give further explanation...

aliculPix4D (approved):
Pre-approving...

I left some questions mostly for my understanding...

In general, I like the new retry pkg quite a lot.. What was confusing a bit during the review was the fact that you need to import both Github and retry pkgs in pipeline setter... I understand why... it just took me a bit of time to connect the dots.

cogito/ghcommitsink.go
marco-m-pix4d: Cogito uses hashicorp/go-hclog to have structured logging.
Since go 1.21, the stdlib has structured logging also, in the log/slog package.
Thus, there is no reason anymore to use hashicorp/go-hclog.

In particular, package cogito/retry is independent from the rest of cogito and is meant to be imported by other tools.

So, if we were to keep using go-hclog also for cogito/retry, we would be imposing go-hclog as a dependency on any tool importing cogito/retry.

To avoid rewriting the whole cogito to use log/slog in this PR, we are using an adapter from go-hclog to log/slog, that allows both to transparently cohexist.

internal/hclogslog/adapter.go
marco-m-pix4d: See also explanation in previous comment about log/slog and the adapter.

aliculPix4D: > If you explain why you are not a fan,

I think this is what is bothering me:
- inconsistent use of multiple pkgs to achieve the same thing. The same is true with my comment about the new imported test package.
- complexity of this new code: i need to understand this new code, just because I am adapting to the fact in point 1

> I can give further explanation...

no need to bother with this... just venting a bit... I just have the feeling we are importing two much stuff, that can be simplified and make understanding overall the cogito package unnecessarily more complex...

cogito/ghcommitsink.go
aliculPix4D: > To avoid rewriting the whole cogito to use log/slog in this PR

I didn't say I don't understand the reasoning, I just said I don't like it... I would prefer if this is not kept for a longer period of time.

> we would be imposing go-hclog as a dependency on any tool importing cogito/retry.

Maybe that was not such a bad option to keep the focus on the important stuff in this PR, which is the retry mechanism.
Now its to late for that and I don't want or expect any changes regarding this topic, just trying to understand better some decisions you took.

Good things: 
- I understood the reasoning even before your comment
- Personally, I also don't like unnecessary importing of dependencies and this PR will avoid that for the end user of the retry package.
Bad things: 
- I am in dilemma here: would doing this in steps be such a bad idea? first use `go-hclog ` in retry pkg, then switch everything...

retry/retry.go
marco-m-pix4d: This is really at the crux of the Retry API.

The documentation and signature for ClassifierFunc is:
```
// ClassifierFunc decides whether to proceed or not; called by [Retry.Do]
type ClassifierFunc func(err error) Action
```

It takes an `error`, classifies it and returns an `Action`.
The fact that it returns `Success` does _not_ imply that `err == nil`; it only implies that `err`, for the _specific_ use case for which `Retry` is used, represents a success.

Although we can probably state that, if `err ==nil`, then _any_ classifier will return `Success`.

This is why at line 79 we return err: because 99% of the cases it will be nil, but not always.

marco-m-pix4d: This approach is the key to allow `Retry` to be used for _any_ situation, without imposing on it any dependency or knowledge of certain types...

This idea was already present in https://github.com/eapache/go-resiliency/tree/main/retrier; the biggest change I made was to make function BackoffFunc more complicated :-D

retry/retry_test.go
marco-m-pix4d: Good question. In my experiments I found "github.com/go-quicktest/qt" better than (the already good) "gotest.tools/v3/assert", for two reasons: it uses generics, so some misuses are caught at compile time, and it prints clearer diffs in certain cases.

You are right that we should be using only one (in the same spirit of log/slog). I will do the replacement, but as a subsequent PR.

cogito/ghcommitsink.go
marco-m-pix4d: In general I agree that a PR should stay as focused as possible. Now, from my point of view, in this case the focus of the PR is not impacted, because the only package using log/slog is cogito/retry, which did not exist before this PR.

On the other hand, the argument I can make for the introduction of go-quicktest/qt is less strong. Although still limited to the new package cogito/retry (so not adding confusing, unrelated diffs to the PR), go-quicktest/qt is also an additional dependency for cogito (until we replace it everywhere).

cogito/ghcommitsink.go
marco-m-pix4d: fixed.

github/commitstatus_private_test.go
marco-m-pix4d: Deleted. These tests are replaced by
- retry/retry_test.go
- github/commitstatus_test.go

internal/hclogslog/adapter.go
marco-m-pix4d: Created some tickets in PCI backlog to cover this.

retry/retry.go
marco-m-pix4d: @aliculPix4D 

There are two related aspects that I don't like:
- The current code is giving to the user a bare struct, `Retry`, for which the "zero values" of the fields are not usable.
- This more or less works because the only method of Retry is `Do`, which (exactly here in this piece of code) does the role of a constructor.

We have two options here:

1. Make the zero values usable. In this case, this would mean to use the zero values as an hint that the user wants the **default** values. So we would have to come up with reasonable default values (cannot be 0) for `FirstDelay` and `BackoffLimit`. On the other hand, the default value for `SleepFn` is clear: it is the real sleep, `time.Sleep`.
2. Introduce a `New` "constructor" that returns an error, instead of returning the error here in `Do`. In general, "fail fast".

marco-m-pix4d: For details on the concept "make the zero value useful" (to which I would add: "when it is possible"), see https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful.

In general, we can argue that the lack of proper constructors in Go is more a defect than a merit, as this whole discussion indirectly shows...

retry/retry_example_test.go
marco-m-pix4d: Ah, I think I understand what you mean. You can enter here if in line 28 above, `// Do work...`, something fails and we return an error. I can make it clearer there.

aliculPix4D: Yes, but since the function `workFn` here always returns  `nil` you can't enter this if statement in this case...

marco-m-pix4d: Maybe I was not clear:

A comment of the type `// Do work...` is a placeholder for real code. I now extended that comment to make it clearer that it can return an error. I cannot really put a `return err` because it would break the test.

The fact that the literal code in the example does not contain a `return err`, does not imply that the classifier should not check for an error. The classifier MUST check for errors, it is the whole reason of existence of the classifier function...

Now, if this is not clear to you as a reader, it means that I must change the example. I can make it so it returns an error the first time or something similar. Do you feel that we need something like that?

aliculPix4D: > The fact that the literal code in the example does not contain a return err

yes this was the confusing part: literal code missing this

> I can make it so it returns an error the first time or something similar.

Personally, I think it would make it clearer, but do as you think its best. Probably adding that comment is enough to clarify. On first read, my expectation was to see the literal code, since that was not there, I found it confusing later why we are checking for something that can't be achieved.

marco-m-pix4d: I will have a look at this, but after the release.

PR #143, opened 2023-10-19
PCI-3363 add support for Github Enterprise API endpoint

aliculPix4D:
Based on: https://github.com/Pix4D/cogito/pull/139

The first commit is a squashed commit of all commits present in the above PR. Credits are given to both authors: @wanderanimrod and @dev-jin

The second commit simplifies a bit their work and it's basically a cleanup of the `GhAPI` field from top-level code. There was no need to keep using both `GhAPI` and `GithubApiEndpoint`. 

While on it, I tried cleanup a bit the helper functions to prevent the unwanted bugs due to hardcoding `github.com` domain.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v0.12.0] - Unreleased
+
+### Added 
+
+- `source.github_api_endpoint`: override the default API endpoint. This allows you to post commit statuses to repositories hosted by GitHub Enterprise (GHE) instances. When provided, this parameter must be a valid HTTPS URL.
+
 ## [v0.11.0] - 2023-09-22
 
 ### Added 
--- a/README.md
+++ b/README.md
@@ -198,6 +198,10 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
   The log level (one of `debug`, `info`, `warn`, `error`, `silent`).\
   Default: `info`.
 
+- `github_api_endpoint`:\
+  Override the default API endpoint. This allows you to post commit statuses to repositories hosted by GitHub Enterprise (GHE) instances. When provided, this parameter must be a valid https url.
+  Default: `https://api.github.com`
+
 - `log_url`. **DEPRECATED, no-op, will be removed**\
   A Google Hangout Chat webhook. Useful to obtain logging for the `check` step for Concourse < v7.x
 
--- a/cmd/cogito/main.go
+++ b/cmd/cogito/main.go
@@ -12,7 +12,6 @@ import (
 	"github.com/hashicorp/go-hclog"
 
 	"github.com/Pix4D/cogito/cogito"
-	"github.com/Pix4D/cogito/github"
 	"github.com/Pix4D/cogito/sets"
 )
 
@@ -56,7 +55,7 @@ func mainErr(in io.Reader, out io.Writer, logOut io.Writer, args []string) error
 	case "in":
 		return cogito.Get(log, input, out, args[1:])
 	case "out":
-		putter := cogito.NewPutter(github.API, log)
+		putter := cogito.NewPutter(log)
 		return cogito.Put(input, out, args[1:], putter)
 	default:
 		return fmt.Errorf("cli wiring error; please report")
--- a/cmd/cogito/main.go
+++ b/cmd/cogito/main.go
@@ -50,21 +50,14 @@ func mainErr(in io.Reader, out io.Writer, logOut io.Writer, args []string) error
 	})
 	log.Info(cogito.BuildInfo())
 
-	ghAPI := os.Getenv("COGITO_GITHUB_API")
-	if ghAPI != "" {
-		log.Info("endpoint override", "COGITO_GITHUB_API", ghAPI)
-	} else {
-		ghAPI = github.API
-	}
-
 	switch cmd {
 	case "check":
 		return cogito.Check(log, input, out, args[1:])
 	case "in":
 		return cogito.Get(log, input, out, args[1:])
 	case "out":
-		putter := cogito.NewPutter(ghAPI, log)
-		return cogito.Put(log, input, out, args[1:], putter)
+		putter := cogito.NewPutter(github.API, log)
+		return cogito.Put(input, out, args[1:], putter)
 	default:
 		return fmt.Errorf("cli wiring error; please report")
 	}
--- a/cmd/cogito/main_test.go
+++ b/cmd/cogito/main_test.go
@@ -67,6 +67,11 @@ func TestRunPutSuccess(t *testing.T) {
 	chatReply := googlechat.MessageReply{}
 	var gchatUrl *url.URL
 	googleChatSpy := testhelp.SpyHttpServer(&chatMsg, chatReply, &gchatUrl, http.StatusOK)
+	gitHubSpyDomain, err := url.Parse(gitHubSpy.URL)
+	if err != nil {
+		t.Fatalf("error parsing SpyHttpServer URL: %s", err)
+	}
+
 	in := bytes.NewReader(testhelp.ToJSON(t, cogito.PutRequest{
 		Source: cogito.Source{
 			Owner:             "the-owner",
@@ -81,9 +86,9 @@ func TestRunPutSuccess(t *testing.T) {
 	var out bytes.Buffer
 	var logOut bytes.Buffer
 	inputDir := testhelp.MakeGitRepoFromTestdata(t, "../../cogito/testdata/one-repo/a-repo",
-		testhelp.HttpsRemote("the-owner", "the-repo"), "dummySHA", wantGitRef)
+		testhelp.HttpsRemote(gitHubSpyDomain.Host, "the-owner", "the-repo"), "dummySHA", wantGitRef)
 
-	err := mainErr(in, &out, &logOut, []string{"out", inputDir})
+	err = mainErr(in, &out, &logOut, []string{"out", inputDir})
 
 	assert.NilError(t, err, "\nout: %s\nlogOut: %s", out.String(), logOut.String())
 	//
@@ -115,7 +120,7 @@ func TestRunPutSuccessIntegration(t *testing.T) {
 	var out bytes.Buffer
 	var logOut bytes.Buffer
 	inputDir := testhelp.MakeGitRepoFromTestdata(t, "../../cogito/testdata/one-repo/a-repo",
-		testhelp.HttpsRemote(gitHubCfg.Owner, gitHubCfg.Repo), gitHubCfg.SHA,
+		testhelp.HttpsRemote("github.com", gitHubCfg.Owner, gitHubCfg.Repo), gitHubCfg.SHA,
 		"ref: refs/heads/a-branch-FIXME")
 	t.Setenv("BUILD_JOB_NAME", "TestRunPutSuccessIntegration")
 	t.Setenv("ATC_EXTERNAL_URL", "https://cogito.invalid")
--- a/cmd/cogito/main_test.go
+++ b/cmd/cogito/main_test.go
@@ -69,11 +69,12 @@ func TestRunPutSuccess(t *testing.T) {
 	googleChatSpy := testhelp.SpyHttpServer(&chatMsg, chatReply, &gchatUrl, http.StatusOK)
 	in := bytes.NewReader(testhelp.ToJSON(t, cogito.PutRequest{
 		Source: cogito.Source{
-			Owner:        "the-owner",
-			Repo:         "the-repo",
-			AccessToken:  "the-secret",
-			GChatWebHook: googleChatSpy.URL,
-			LogLevel:     "debug",
+			Owner:             "the-owner",
+			Repo:              "the-repo",
+			AccessToken:       "the-secret",
+			GithubApiEndpoint: gitHubSpy.URL,
+			GChatWebHook:      googleChatSpy.URL,
+			LogLevel:          "debug",
 		},
 		Params: cogito.PutParams{State: wantState},
 	}))
@@ -81,7 +82,6 @@ func TestRunPutSuccess(t *testing.T) {
 	var logOut bytes.Buffer
 	inputDir := testhelp.MakeGitRepoFromTestdata(t, "../../cogito/testdata/one-repo/a-repo",
 		testhelp.HttpsRemote("the-owner", "the-repo"), "dummySHA", wantGitRef)
-	t.Setenv("COGITO_GITHUB_API", gitHubSpy.URL)
 
 	err := mainErr(in, &out, &logOut, []string{"out", inputDir})
 
--- a/cogito/protocol.go
+++ b/cogito/protocol.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/Pix4D/cogito/github"
 	"github.com/Pix4D/cogito/sets"
 )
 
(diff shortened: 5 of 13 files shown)


cogito/putter.go
```diff
@@ -262,23 +259,24 @@ func checkGitRepoDir(dir, owner, repo string) error {
 	if err != nil {
 		return fmt.Errorf(".git/config: remote: %w", err)
 	}
-	left := []string{"github.com", owner, repo}
```
aliculPix4D, line 265: So as discussed, for this PR we will not validate the domain, but I still want to give the endpoint information to the end user, just in case. So even though we don't strictly compare the `gu.URL.Host` with the configured github_api_endpoint we at least print both. So at least this will simplify the debugging a bit. 

For me this is still an open question, but we will discuss it when @marco-m-pix4d is back. For now I go with this beacuse:
- It its how the authors did it in the first commit
- @marco-m-pix4d recommended to ease the validation for now: https://github.com/Pix4D/cogito/pull/142#discussion_r1364192679

aliculPix4D: I see only two options here: 
- either we ease the validation as its done here
- or we support configuring both `github_endpoint` and `github_api_endpoint`. 

How would this work?
- if both are unset in source config: `github_endpoint` points to: `https://github.com` and `github_api_endpoint` points to: `https://api.github.com`
- if only `github_endpoint` is set.  `github_api_endpoint` points to: `https://<github_endpoint>/api/v3/`. This is the default API endpoint for GitHub Enterprise in my understanding
- if both are set we use the values that are set

This would enable us to due this validation using the configured: `github_endpoint` and do the API request to `github_api_endpoint`.

@wanderanimrod Opinions on the alternative solution?

cmd/cogito/main.go
```diff
@@ -50,21 +49,14 @@ func mainErr(in io.Reader, out io.Writer, logOut io.Writer, args []string) error
...
-	}
-
 	switch cmd {
 	case "check":
 		return cogito.Check(log, input, out, args[1:])
 	case "in":
 		return cogito.Get(log, input, out, args[1:])
 	case "out":
-		putter := cogito.NewPutter(ghAPI, log)
-		return cogito.Put(log, input, out, args[1:], putter)
```
aliculPix4D, line 67: `log` variable was really unused. There is another option where we can not set the log in putter structure, but send it always as a function parameter, but that is too painful in my opinion. Too many code changes for no gain.

cogito/protocol_test.go
```diff
@@ -102,6 +110,36 @@ func TestSourceValidationFailure(t *testing.T) {
...
+		},
+		{
+			name: "invalid http protocol prefix in git source github_api_endpoint",
+			source: cogito.Source{
+				Owner:             "the-owner",
+				Repo:              "the-repo",
+				AccessToken:       "the-token",
+				GithubApiEndpoint: "john.smith.cim",
+			},
+			wantErr: "source: github_api_endpoint 'john.smith.cim' is an invalid api endpoint",
```
odormond, line 141: First of all, I don't like much the wording "protocol prefix". There is no prefix to a protocol.
Second, I don't understand why you have the third test. It should fail for the same reason the first one fails: no "protocol prefix".

Now, let's try to come up with a better wording. You're expecting the `GithubApiEndpoint` to hold a URI for the `https` _scheme_. The https scheme expect an _authority_ that must be introduced by `//`. After that their might be a _path_ that will start at first `/` after the authority.

https://datatracker.ietf.org/doc/html/rfc3986.html#section-3

cmd/cogito/main_test.go
```diff
@@ -67,23 +67,28 @@ func TestRunPutSuccess(t *testing.T) {
 	chatReply := googlechat.MessageReply{}
 	var gchatUrl *url.URL
 	googleChatSpy := testhelp.SpyHttpServer(&chatMsg, chatReply, &gchatUrl, http.StatusOK)
+	gitHubSpyDomain, err := url.Parse(gitHubSpy.URL)
```
aliculPix4D, line 70: maybe I should move this part closer to where its used; around  L88 ?

cogito/protocol_test.go
aliculPix4D: Yes, you are correct both first and third test, return the same error.
```
parse "github.coffee.com/api/v3": invalid URI for request
parse "john.smith.cim": invalid URI for request
```
I will remove the third one and adapt the wording according to the naming convention in the test name. Thanks

aliculPix4D: Done as suggested.

odormond (approved):

wanderanimrod:
@aliculPix4D , @dev-jin has been working with me on this. The tests on our end are complete.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v0.12.0] - Unreleased
+
+### Added 
+
+- `source.github_api_endpoint`: override the default API endpoint. This allows you to post commit statuses to repositories hosted by GitHub Enterprise (GHE) instances. When provided, this parameter must be a valid HTTPS URL.
```
marco-m-pix4d, line 12: I would make it more evident that the change adds support for GH Enterprise and that the initial contribution comes from XXX as follows:

```
- Support GitHub Enterprise (see `source.github_api_endpoint` in the README). Initial implementation by @wanderanimrod and @dev-jin.
```

README.md
```diff
@@ -198,6 +198,10 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
   The log level (one of `debug`, `info`, `warn`, `error`, `silent`).\
   Default: `info`.
 
+- `github_api_endpoint`:\
```
marco-m-pix4d, line 201: The term "endpoint" means a specific URL (each API has N endpoints, all on the same hostname), so here it would be misleading.

To find a more precise name, we could use the same terminology as the GH cli tool `gh`:

```
$ gh help auth login
Authenticate with a GitHub host.
...
FLAGS
  -h, --hostname string       The hostname of the GitHub instance to authenticate with
...
EXAMPLES
  # authenticate with a specific GitHub instance
  $ gh auth login --hostname enterprise.internal
```

So we could call this parameter `github_hostname`.

README.md
```diff
@@ -198,6 +198,10 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
...
+  Override the default API endpoint. This allows you to post commit statuses to repositories hosted by GitHub Enterprise (GHE) instances. When provided, this parameter must be a valid https url.
```
marco-m-pix4d, line 202: For the same reasoning as above, "Override the default API endpoint" -> "Override the default API hostname" or similar.

cmd/cogito/main_test.go
```diff
@@ -69,21 +69,25 @@ func TestRunPutSuccess(t *testing.T) {
...
 		Source: cogito.Source{
-			Owner:        "the-owner",
-			Repo:         "the-repo",
-			AccessToken:  "the-secret",
-			GChatWebHook: googleChatSpy.URL,
-			LogLevel:     "debug",
+			Owner:             "the-owner",
+			Repo:              "the-repo",
+			AccessToken:       "the-secret",
+			GithubApiEndpoint: gitHubSpy.URL,
```
marco-m-pix4d, line 75: Similarly as a previous comment, here we need a name like GitHubApiHostname, without the "endpoint" part.

cmd/cogito/main.go
aliculPix4D: > I am surprised here, because I remember we had to set the log here (see line 45) because of line 41, were we "peek" for the log level before parsing the JSON of the input.

Yes, sure. that is still the case. There is no changes to that "logic"

> Ah. Do you mean that the putter (line 58) still uses the log initialized here, while cogito.Put (line 59) currently does not log anything?

Exactly.

> Also in this case, I would be tempted to leave the log parameter to cogito.Put, because it matches the two other calls, cogito.Check and cogito.Get. One less difference.

Apparently, some linters complain about unused parameters. Our doesn't. I did the exercise and tried to make it consistent with other two calls, but its not worth it really. So we need a decision on one of the options:
- we live with unused parameter that bring the confusion (belive me it does) just to have a match with above calls
- or we remove it

cogito/protocol.go
```diff
@@ -268,6 +277,9 @@ func (src *Source) Validate() error {
 	if len(src.ChatNotifyOnStates) == 0 {
 		src.ChatNotifyOnStates = defaultNotifyStates
 	}
+	if len(src.GithubApiEndpoint) == 0 {
```
marco-m-pix4d, line 280: The parameter is a string, so I would expect `if src.GithubApiEndpoint == ""`, and is also what we use at line 264 above...

cogito/protocol.go
```diff
@@ -257,7 +261,12 @@ func (src *Source) Validate() error {
 	//
 	// Validate optional fields.
 	//
-	// In this case, nothing to validate.
+	if src.GithubApiEndpoint != "" {
+		u, err := url.ParseRequestURI(src.GithubApiEndpoint)
+		if err != nil || u.Host == "" {
+			return fmt.Errorf("source: github_api_endpoint '%s' is an invalid api endpoint", src.GithubApiEndpoint)
```
marco-m-pix4d, line 267: Here I have the impression that we are removing details from the error that we actually have and would help the user, or do we have a particular reason?

```go
if src.GithubApiEndpoint != "" {
    u, err := url.ParseRequestURI(src.GithubApiEndpoint)
    if err != nil {
        return fmt.Errorf("source: invalid github_api_endpoint %q: %s", src.GithubApiEndpoint, err)
    }
    if u.Host == "" {
        return fmt.Errorf("source: invalid github_api_endpoint %q: %s", src.GithubApiEndpoint, "empty hostname")
    }
}
```

marco-m-pix4d (commented):
@aliculPix4D I left some comments inline. Could you please have a look and ping me if you have any question? Regarding your questions for cogito/putter.go ("I still want to give the endpoint information to the end user, ...") I think it is better if we go through them via screen sharing, please ping me when you are ready.

CHANGELOG.md
aliculPix4D: Changed as suggested.

README.md
aliculPix4D: Changed as suggested.

README.md
aliculPix4D: Changed as suggested.

cmd/cogito/main_test.go
aliculPix4D: Changed as suggested.

cogito/protocol.go
aliculPix4D: Changed as suggested.

cogito/protocol.go
aliculPix4D: Changed as suggested.

cogito/putter.go
```diff
@@ -262,23 +259,24 @@ func checkGitRepoDir(dir, owner, repo string) error {
...
+    repo: %s
 
 Cogito SOURCE configuration:
+    github_api_hostname: %s
     owner: %s
-     repo: %s`,
+    repo: %s`,
 				gitUrl, gu.Owner, gu.Repo,
-				owner, repo)
+				endpoint, owner, repo)
```
aliculPix4D, line 279: change the name: endpoint --> hostname

cogito/putter.go
aliculPix4D: @marco-m-pix4d see the 3 possibilities:
https://github.com/google/go-github/blob/66ed84deba8936684a4a6bb09b00135f326064fc/github/github.go#L363

and the naming convention they use.

cmd/cogito/main.go
marco-m-pix4d: (discussed off-line)

cogito/putter.go
marco-m-pix4d: @aliculPix4D thanks for the explanation.

Looking at the official documentation for GH Enterprise (https://docs.github.com/en/enterprise-server@3.10/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28), I could not find, as you mentioned, the fact that if the hostname contains the `api` subdomain, then the URL should not contain "/api/v3": it seems that all GH Enterprise installation want (or support) hostname/api/v3 ...

I then though that instead of considering `go-github` as authoritative (it comes from Google, not GibHub), we could maybe look at the `gh` tool (which does come from GitHub). There I found https://github.com/cli/cli/blob/trunk/internal/ghinstance/host.go, which seems to contain all the logic to make a decision. Have a look also at the test file, https://github.com/cli/cli/blob/20baccfa85d15963eb1ab4f750da0da37b0af7f5/internal/ghinstance/host_test.go.

There, in function `RESTPrefix` https://github.com/cli/cli/blob/20baccfa85d15963eb1ab4f750da0da37b0af7f5/internal/ghinstance/host.go#L85-L96, they seem to have all the logic:
```go
func RESTPrefix(hostname string) string {
	if isGarage(hostname) {
		return fmt.Sprintf("https://%s/api/v3/", hostname)
	}
	if IsEnterprise(hostname) {
		return fmt.Sprintf("https://%s/api/v3/", hostname)
	}
	if strings.EqualFold(hostname, localhost) {
		return fmt.Sprintf("http://api.%s/", hostname)
	}
	return fmt.Sprintf("https://api.%s/", hostname)
}
```
from that, it seems that there are only the following possibilities:
```
garage.github.com                    -> garage.github.com/api/v3/
anything not github.com (enterprise) -> anything/api/v3/
github.localhost                     -> api.github.localhost
github.com                           -> api.github.com (this I think)
```

Now, as we know, anything below `internal` means that it can break at any moment, but it still seems reasonable to me. It is similar to the logic in `go-github`. Maybe we could lift the whole file and be done with it (license is MIT) ?

Going back to your proposal:

> or we support configuring both `github_endpoint` and `github_api_endpoint`.

Do I understand correctly that the reason to introduce 2 parameters is to be able to use directly `github_endpoint` to compare with the git remote from `.git/config` without having to remove the `/api/v3` path prefix?

The reason I am going slowly is to see if we can both keep the configuration as simple as possible and avoid painting ourselves in a corner if we make an error.

aliculPix4D: > Do I understand correctly that the reason to introduce 2 parameters is to be able to use directly github_endpoint to compare with the git remote from .git/config without having to remove the /api/v3 path prefix?

Yes, but given your findings that is not needed.

As I see now, we should support setting only: `github_hostname`, and default is: `github.com`, and then we should add the `api.` prefix for default or `/api/v3` suffix to the `github_hostname` when API url is required.

This will allow for simple validation  against the git remote from .git/config and using the correct API url when required.

cmd/cogito/main.go
aliculPix4D: This change is reverted in:
https://github.com/Pix4D/cogito/pull/143/commits/95d29b29c448c536ae77895298f4d44887a83b06

aliculPix4D:
@marco-m-pix4d just for your info: sadly I was correct. There is an implicit relationship also with Gchat putter.
https://github.com/Pix4D/cogito/blob/9558fa3fe5454cc1e1ff44c144cef9786862485b/cogito/gchatsink.go#L125
and the value is hardcoded there. This keeps getting better and better.

marco-m-pix4d:
@aliculPix4D 

> sadly I was correct. There is an implicit relationship also with Gchat putter.

Good catch.

As mentioned face to face, we can split this in two. First we focus on delivering cogito for Enterprise GitHub. The release notes will mention in section "Known bugs" that the URL in the gchat message will be wrong.

 Then, as a separate ticket, we make the gchat message correct also for Enterprise GitHub.

aliculPix4D:
Closing in favor of: https://github.com/Pix4D/cogito/pull/143

SpectreVert:
> Closing in favor of: #143

#146 no?

aliculPix4D:
> > Closing in favor of: #143
> 
> #146 no?

Yes, it was wrong, but more precisely: https://github.com/Pix4D/cogito/pull/144 and https://github.com/Pix4D/cogito/pull/145

PR #144, opened 2023-10-27
PCI-3363 don't hardcode GitHub domain name

aliculPix4D:
- adds a possibility to configure `source.github_hostname` to allow for overriding of the GitHub hostname only from the tests. This helps us avoid a very ugly implementation with: `os.Getenv("COGITO_GITHUB_API")`. Default value is of course: `github.com`
- this step forces us to remove the hardcoded `github.com` from basically everywhere
- even the implicit dependency in Gchat putter that I reported is fixed (https://github.com/Pix4D/cogito/commit/362de679d8ba054f909d972f31b4796268f921bd)

We then later craft the API endpoint using: `func apiEndpoint(h string) string` function.

No tests were added, at least not for now. I think everything is covered by the existing tests. I could add the unit test for `apiEndpoint` function actually.

I will open a draft follow-up PR on top of this one, just to show you how simple will be to allow the GH Enterprise if this PR is merged.

### DIFF
--- a/cmd/cogito/main_test.go
+++ b/cmd/cogito/main_test.go
@@ -80,7 +80,7 @@ func TestRunPutSuccess(t *testing.T) {
 	var out bytes.Buffer
 	var logOut bytes.Buffer
 	inputDir := testhelp.MakeGitRepoFromTestdata(t, "../../cogito/testdata/one-repo/a-repo",
-		testhelp.HttpsRemote("the-owner", "the-repo"), "dummySHA", wantGitRef)
+		testhelp.HttpsRemote("github.com", "the-owner", "the-repo"), "dummySHA", wantGitRef)
 	t.Setenv("COGITO_GITHUB_API", gitHubSpy.URL)
 
 	err := mainErr(in, &out, &logOut, []string{"out", inputDir})
@@ -115,7 +115,7 @@ func TestRunPutSuccessIntegration(t *testing.T) {
 	var out bytes.Buffer
 	var logOut bytes.Buffer
 	inputDir := testhelp.MakeGitRepoFromTestdata(t, "../../cogito/testdata/one-repo/a-repo",
-		testhelp.HttpsRemote(gitHubCfg.Owner, gitHubCfg.Repo), gitHubCfg.SHA,
+		testhelp.HttpsRemote("github.com", gitHubCfg.Owner, gitHubCfg.Repo), gitHubCfg.SHA,
 		"ref: refs/heads/a-branch-FIXME")
 	t.Setenv("BUILD_JOB_NAME", "TestRunPutSuccessIntegration")
 	t.Setenv("ATC_EXTERNAL_URL", "https://cogito.invalid")
--- a/cmd/cogito/main_test.go
+++ b/cmd/cogito/main_test.go
@@ -63,6 +63,10 @@ func TestRunPutSuccess(t *testing.T) {
 	var ghReq github.AddRequest
 	var ghUrl *url.URL
 	gitHubSpy := testhelp.SpyHttpServer(&ghReq, nil, &ghUrl, http.StatusCreated)
+	gitHubSpyURL, err := url.Parse(gitHubSpy.URL)
+	if err != nil {
+		t.Fatalf("error parsing SpyHttpServer URL: %s", err)
+	}
 	var chatMsg googlechat.BasicMessage
 	chatReply := googlechat.MessageReply{}
 	var gchatUrl *url.URL
@@ -72,6 +76,7 @@ func TestRunPutSuccess(t *testing.T) {
 			Owner:        "the-owner",
 			Repo:         "the-repo",
 			AccessToken:  "the-secret",
+			GhHostname:   gitHubSpyURL.Host,
 			GChatWebHook: googleChatSpy.URL,
 			LogLevel:     "debug",
 		},
@@ -80,9 +85,9 @@ func TestRunPutSuccess(t *testing.T) {
 	var out bytes.Buffer
 	var logOut bytes.Buffer
 	inputDir := testhelp.MakeGitRepoFromTestdata(t, "../../cogito/testdata/one-repo/a-repo",
-		testhelp.HttpsRemote("github.com", "the-owner", "the-repo"), "dummySHA", wantGitRef)
+		testhelp.HttpsRemote(gitHubSpyURL.Host, "the-owner", "the-repo"), "dummySHA", wantGitRef)
 
-	err := mainErr(in, &out, &logOut, []string{"out", inputDir})
+	err = mainErr(in, &out, &logOut, []string{"out", inputDir})
 
 	assert.NilError(t, err, "\nout: %s\nlogOut: %s", out.String(), logOut.String())
 	//
@@ -114,7 +119,7 @@ func TestRunPutSuccessIntegration(t *testing.T) {
 	var out bytes.Buffer
 	var logOut bytes.Buffer
 	inputDir := testhelp.MakeGitRepoFromTestdata(t, "../../cogito/testdata/one-repo/a-repo",
-		testhelp.HttpsRemote("github.com", gitHubCfg.Owner, gitHubCfg.Repo), gitHubCfg.SHA,
+		testhelp.HttpsRemote(github.GhDefaultHostname, gitHubCfg.Owner, gitHubCfg.Repo), gitHubCfg.SHA,
 		"ref: refs/heads/a-branch-FIXME")
 	t.Setenv("BUILD_JOB_NAME", "TestRunPutSuccessIntegration")
 	t.Setenv("ATC_EXTERNAL_URL", "https://cogito.invalid")
--- a/cmd/cogito/main_test.go
+++ b/cmd/cogito/main_test.go
@@ -81,7 +81,6 @@ func TestRunPutSuccess(t *testing.T) {
 	var logOut bytes.Buffer
 	inputDir := testhelp.MakeGitRepoFromTestdata(t, "../../cogito/testdata/one-repo/a-repo",
 		testhelp.HttpsRemote("github.com", "the-owner", "the-repo"), "dummySHA", wantGitRef)
-	t.Setenv("COGITO_GITHUB_API", gitHubSpy.URL)
 
 	err := mainErr(in, &out, &logOut, []string{"out", inputDir})
 
--- a/cmd/cogito/main_test.go
+++ b/cmd/cogito/main_test.go
@@ -64,9 +64,7 @@ func TestRunPutSuccess(t *testing.T) {
 	var ghUrl *url.URL
 	gitHubSpy := testhelp.SpyHttpServer(&ghReq, nil, &ghUrl, http.StatusCreated)
 	gitHubSpyURL, err := url.Parse(gitHubSpy.URL)
-	if err != nil {
-		t.Fatalf("error parsing SpyHttpServer URL: %s", err)
-	}
+	assert.NilError(t, err, "error parsing SpyHttpServer URL: %s", err)
 	var chatMsg googlechat.BasicMessage
 	chatReply := googlechat.MessageReply{}
 	var gchatUrl *url.URL
--- a/cogito/ghcommitsink.go
+++ b/cogito/ghcommitsink.go
@@ -1,9 +1,7 @@
 package cogito
 
 import (
-	"fmt"
 	"log/slog"
-	"strings"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
@@ -43,7 +41,7 @@ func (sink GitHubCommitStatusSink) Send() error {
 	context := ghMakeContext(sink.Request)
 
 	target := &github.Target{
-		Server: apiEndpoint(sink.Request.Source.GhHostname),
+		Server: github.ApiRoot(sink.Request.Source.GhHostname),
 		Retry: retry.Retry{
 			FirstDelay:   retryFirstDelay,
 			BackoffLimit: retryBackoffLimit,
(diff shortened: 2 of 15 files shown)


cogito/ghcommitsink.go
```diff
@@ -90,3 +91,11 @@ func ghMakeContext(request PutRequest) string {
 	}
 	return context
 }
+
+func apiEndpoint(h string) string {
```
aliculPix4D, line 95: maybe the better place would be to move this functions to github pkg?

testhelp/testhelper.go
```diff
@@ -247,18 +247,18 @@ func MakeGitRepoFromTestdata(
...
-	return fmt.Sprintf("https://github.com/%s/%s.git", owner, repo)
+func HttpsRemote(domain, owner, repo string) string {
+	return fmt.Sprintf("https://%s/%s/%s.git", domain, owner, repo)
 }
 
 // HttpRemote returns a GitHub HTTP URL
-func HttpRemote(owner, repo string) string {
-	return fmt.Sprintf("http://github.com/%s/%s.git", owner, repo)
+func HttpRemote(domain, owner, repo string) string {
+	return fmt.Sprintf("http://%s/%s/%s.git", domain, owner, repo)
```
odormond, line 261: Why call it `hostname` everywhere else and here use `domain` instead?

odormond (approved):

testhelp/testhelper.go
marco-m-pix4d: +1

cmd/cogito/main_test.go
```diff
@@ -63,6 +63,10 @@ func TestRunPutSuccess(t *testing.T) {
 	var ghReq github.AddRequest
 	var ghUrl *url.URL
 	gitHubSpy := testhelp.SpyHttpServer(&ghReq, nil, &ghUrl, http.StatusCreated)
+	gitHubSpyURL, err := url.Parse(gitHubSpy.URL)
+	if err != nil {
+		t.Fatalf("error parsing SpyHttpServer URL: %s", err)
+	}
```
marco-m-pix4d, line 69: we can `assert.NilError` here as usual

cogito/ghcommitsink.go
marco-m-pix4d: Yes it should go in the github package.
Note also that the name `apiEndpoint` is wrong, this is not returning an endpoint, it is returning the root or prefix of all the endpoints. Considering it being in package github, it could be something like

`github.ApiPrefix`
or
`github.ApiRoot`

I tend to think that `ApiRoot` is better ...

marco-m-pix4d: Ah. I am now confused, I was expecting this function to append also the `/api/v3` (or whatever it was) prefix.

marco-m-pix4d: What is the intent of this function, compared to where we add the `/api/v3` prefix? Can it be merged into one?

Note also that
```
	if strings.Contains(hostname, "127.0.0.1") {
		return fmt.Sprintf("http://%s", hostname)
	}
```
can fail in spectacular ways...

aliculPix4D: This will arrive in the next PR... The next PR should really introduce the support for GH Enterprise (see the PR description).

I separated them to fix first the existing issues, and later concentrate on GH Enterprise feature only.

cogito/ghcommitsink_test.go
```diff
@@ -23,17 +23,21 @@ func TestSinkGitHubCommitStatusSendSuccess(t *testing.T) {
 	var ghReq github.AddRequest
 	var URL *url.URL
 	ts := testhelp.SpyHttpServer(&ghReq, nil, &URL, http.StatusCreated)
+	gitHubSpyURL, err := url.Parse(ts.URL)
+	if err != nil {
+		t.Fatalf("error parsing SpyHttpServer URL: %s", err)
```
marco-m-pix4d, line 28: same comment about `assert.`

cogito/ghcommitsink_test.go
```diff
@@ -47,17 +51,21 @@ func TestSinkGitHubCommitStatusSendFailure(t *testing.T) {
 		http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 			w.WriteHeader(http.StatusTeapot)
 		}))
+	gitHubSpyURL, err := url.Parse(ts.URL)
+	if err != nil {
+		t.Fatalf("error parsing SpyHttpServer URL: %s", err)
```
marco-m-pix4d, line 56: same

cogito/putter_private_test.go
```diff
@@ -143,32 +146,34 @@ func TestCheckGitRepoDirFailure(t *testing.T) {
...
-			repoURL: testhelp.HttpsRemote("owner-a", "repo-a"),
+			repoURL: testhelp.HttpsRemote("github.com", "owner-a", "repo-a"),
 			wantErrWild: `the received git repository is incompatible with the Cogito configuration.
 
 Git repository configuration (received as 'inputs:' in this PUT step):
-      url: https://github.com/owner-a/repo-a.git
+    url: https://github.com/owner-a/repo-a.git
     owner: owner-a
-     repo: repo-a
+    repo: repo-a
```
marco-m-pix4d, line 155: Do we know why these whitespace changes are needed? (here and everywhere else in this PR)
I have a recollection that there are tabs in the output, if this is correct, then maybe all these changes are not needed if we fix where we create the output?

github/commitstatus.go
```diff
@@ -31,12 +31,14 @@ func (e *StatusError) Error() string {
 	return fmt.Sprintf("%s\n%s", e.What, e.Details)
 }
 
+// Default GitHub hostname
+const GhDefaultHostname = "github.com"
+
 // API is the GitHub API endpoint.
```
marco-m-pix4d, line 37: // API is the default GitHub API hostname.

github/commitstatus.go
```diff
@@ -31,12 +31,14 @@ func (e *StatusError) Error() string {
 	return fmt.Sprintf("%s\n%s", e.What, e.Details)
 }
 
+// Default GitHub hostname
```
marco-m-pix4d, line 34: // GhDefaultHostname is the default GitHub hostname (used for git but not for the API)

marco-m-pix4d: We begin the comment with the same name as the variable because this is the hint for the Go doc system to associate the comment to the symbol.

marco-m-pix4d:
@aliculPix4D I think we are almost there. I just left some comments on this apiEndpoint function.

cogito/putter_private_test.go
aliculPix4D: > Do we know why these whitespace changes are needed? (here and everywhere else in this PR)

they are not needed; I just wanted to make them consistent. You mentioned that these were tabs but they were not tabs (at least this is my understanding). In my editor I see; 4 spaces, 5 spaces, 6 spaces; so I changed it such that always its 4 spaces.

marco-m-pix4d: OK.

testhelp/testhelper.go
aliculPix4D: Fixed in: https://github.com/Pix4D/cogito/pull/144/commits/66f1cf52187b4780f4fb0775009bab43f2a2155d

github/commitstatus.go
aliculPix4D: Fixed in: https://github.com/Pix4D/cogito/pull/144/commits/66f1cf52187b4780f4fb0775009bab43f2a2155d

github/commitstatus.go
aliculPix4D: Fixed in: https://github.com/Pix4D/cogito/commit/66f1cf52187b4780f4fb0775009bab43f2a2155d

cogito/ghcommitsink_test.go
aliculPix4D: Fixed in: https://github.com/Pix4D/cogito/commit/66f1cf52187b4780f4fb0775009bab43f2a2155d

cogito/ghcommitsink_test.go
aliculPix4D: Fixed in: https://github.com/Pix4D/cogito/commit/66f1cf52187b4780f4fb0775009bab43f2a2155d

cmd/cogito/main_test.go
aliculPix4D: Fixed in: https://github.com/Pix4D/cogito/commit/66f1cf52187b4780f4fb0775009bab43f2a2155d

SpectreVert (approved):

github/commitstatus.go
aliculPix4D: Actually, is this correct?

` API is the default GitHub API hostname`? hostname? actually in this case its not hostname. It contains a protocol part (`https://`) and the API hostname part: `api.github.com`, so more correct would be base url?

cogito/putter_private_test.go
aliculPix4D: No actions were taken on this point. I double-checked again and for me, it seems that the changes are correct because they try to make the spaces consistent (4 spaces everywhere)

cogito/ghcommitsink.go
aliculPix4D: Ah, yes.. Can you also comment a bit more on:

```
if strings.Contains(hostname, "127.0.0.1") {
		return fmt.Sprintf("http://%s", hostname)
	}
```
> can fail in spectacular ways...

As I see it; `127.0.0.1` is only configured from tests if this is a HTTP testserver. What am I missing here? 

In that ghinstance code from gh cli they did:
```
// Localhost is the domain name of a local GitHub instance.
const localhost = "github.localhost"

if strings.EqualFold(hostname, localhost) {
    return fmt.Sprintf("http://api.%s/", hostname)
}
```
since our localhost is always of the format: `127.0.0.1:PORT` I couldn't use the `==`

marco-m-pix4d: Example: `127.0.0.1.hello.com` will match the conditional, and also `foo.127.0.0.1.hello.com`
In this case it is harmless, but in general one needs to be careful when manipulating user input.

aliculPix4D:
@marco-m-pix4d @odormond ready for another round of review: https://github.com/Pix4D/cogito/pull/144/commits/f826246acc64f7f5b4775e000ca2b306d0f1d02a

github/commitstatus.go
```diff
@@ -31,12 +32,16 @@ func (e *StatusError) Error() string {
 	return fmt.Sprintf("%s\n%s", e.What, e.Details)
 }
 
-// API is the GitHub API endpoint.
+// GhDefaultHostname is the default GitHub hostname (used for git but not for the API)
+const GhDefaultHostname = "github.com"
+
+// API is the default GitHub API root URL.
 const API = "https://api.github.com"
```
aliculPix4D, line 39: @marco-m-pix4d I know that I suggested removing this constant and just hardcode it in `ApiRoot`, but I tried and it make 2 tests a bit uglier so I decided to keep it.

Basically one would be required to configure the test server in 2 integration tests like:

```
Source{
  Server: github.ApiRoot(github.GhDefaultHostname)
}
```

compared to what we have now:
```
```
Source{
  Server: github.API
}
```

I don't see any gain with the change I proposed, so I kept it as it is.
```

cogito/ghcommitsink.go
aliculPix4D: Fixed.

github/commitstatus.go
aliculPix4D: Fixed.

cogito/protocol.go
```diff
@@ -257,7 +263,11 @@ func (src *Source) Validate() error {
 	//
 	// Validate optional fields.
 	//
-	// In this case, nothing to validate.
+	if src.GhHostname != "" {
+		if !hostnameRegexp.MatchString(src.GhHostname) {
+			return fmt.Errorf("source: invalid github_api_hostname: %s. Don't configure the schema or the path", src.GhHostname)
+		}
+	}
```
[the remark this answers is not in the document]
aliculPix4D: Not a fan of your suggestion but if that is the team decision, I am ready to do the change.

My point of view:
- for me nested is more readable...
- I got hurt so many times with more complex if statements with multiple conditionals that I try to avoid it... 
 On the other hand, maybe in some simpler cases your approach is justified...

github/commitstatus.go
marco-m-pix4d: There is a gain: only one way to do it. It reduces confusion, since `github.API` doesn't exist, one is forced to use `github.ApiRoot(github.GhDefaultHostname)`, which is what we wanted.

cogito/protocol.go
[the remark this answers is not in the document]
marco-m-pix4d: I think that the reason of Olivier's suggestion is because:
- there is nothing inside the `if` besides the call to hostnameRegexp.MatchString.
- the expression of the `if` is protecting the call of hostnameRegexp.MatchString. A classic would be:
`if ptr != nil && foo(*ptr)`
it is indeed idiomatic to use the && in those cases.

> I got hurt so many times with more complex if statements with multiple conditionals that I try to avoid it...

Very true, this happens to everybody and in any case makes the code easier to misunderstand.

But here we are in the realm of the idiomatic `if ptr != nil && foo(*ptr)`.

Now, there is also another approach that might work in this case: see if we can apply the defaults before arriving here. If we can, then this code would become simply

```
if !hostnameRegexp.MatchString(src.GhHostname) 
```

even simpler and arguably even better.

marco-m-pix4d (approved):

cogito/protocol.go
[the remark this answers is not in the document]
aliculPix4D: Yes, moving this statement after defaults are applied simplifies a lot and additionally verifies that someone doesn't put the wrong value for: `github.GhDefaultHostname`. So win win situation. Thanks to both.

```
if !hostnameRegexp.MatchString(src.GhHostname) {
		return fmt.Errorf("source: invalid github_api_hostname: %s. Don't configure the schema or the path", src.GhHostname)
}
```

PR #145, opened 2023-10-27
PCI-3363 add ghe support

aliculPix4D:
Depends on: https://github.com/Pix4D/cogito/pull/144

- add a new feature (support GH Enterprise)

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,10 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
   The log level (one of `debug`, `info`, `warn`, `error`, `silent`).\
   Default: `info`.
 
+- `github_hostname`:\
+  GitHub hostname. This allows to post commit statuses to repositories hosted by GitHub Enterprise (GHE) instances. For example: github.mycompany.org will be expanded by cogito to https://github.mycompany.org/api/v3.\
+  Default: `github.com`
+
 - `log_url`. **DEPRECATED, no-op, will be removed**\
   A Google Hangout Chat webhook. Useful to obtain logging for the `check` step for Concourse < v7.x
 
--- a/github/commitstatus.go
+++ b/github/commitstatus.go
@@ -196,13 +196,17 @@ func (cs CommitStatus) explainError(err error, state, sha, url string) error {
 
 // ApiRoot constructs the root part of the GitHub API URL for a given hostname.
 // Example:
-// if hostname is: github.com it returns https://api.github.com
-// if hostname is overrriden in tests by localhost HTTP testserver it returns http://127.0.0.1:PORT
+// if hostname is github.com it returns https://api.github.com
+// if hostname looks like a httptest server, it returns http://127.0.0.1:PORT
+// otherwise, hostname is assumed to be of a Github Enterprise instance.
+// For example, github.mycompany.org returns https://github.mycompany.org/api/v3
 func ApiRoot(h string) string {
 	hostname := strings.ToLower(h)
+	if hostname == GhDefaultHostname {
+		return API
+	}
 	if localhostRegexp.MatchString(hostname) {
 		return fmt.Sprintf("http://%s", hostname)
 	}
-
-	return API
+	return fmt.Sprintf("https://%s/api/v3", hostname)
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v0.12.0] - Unreleased
+
+### Added
+
+- Support GitHub Enterprise (see `source.github_hostname` in the README). Initial implementation by @wanderanimrod and @dev-jin.
+
 ## [v0.11.0] - 2023-09-22
 
 ### Added 
--- a/github/commitstatus_test.go
+++ b/github/commitstatus_test.go
@@ -481,6 +481,11 @@ func TestApiRoot(t *testing.T) {
 			hostname: github.GhDefaultHostname,
 			wantAPI:  github.API,
 		},
+		{
+			name:     "Github Enterprise hostname",
+			hostname: "github.mycompany.org",
+			wantAPI:  "https://github.mycompany.org/api/v3",
+		},
 	}
 
 	for _, tc := range testCases {


cogito/ghcommitsink.go
```diff
@@ -93,9 +93,21 @@ func ghMakeContext(request PutRequest) string {
...
+		return github.API
+	}
 	if strings.Contains(hostname, "127.0.0.1") {
 		return fmt.Sprintf("http://%s", hostname)
 	}
-	return github.API
+	return fmt.Sprintf("https://%s/api/v3", hostname)
+}
+
+func normalizeHostname(h string) string {
```
aliculPix4D, line 106: of course, I will rewrite this in a cleaner way; probably using TrimFunction...

cogito/ghcommitsink.go
```diff
@@ -93,9 +93,21 @@ func ghMakeContext(request PutRequest) string {
 }
 
 func apiEndpoint(h string) string {
-	hostname := strings.ToLower(h)
+	hostname := normalizeHostname(h) // not really required; already called during protocol init
+	if hostname == github.GhDefaultHostname {
+		return github.API
+	}
 	if strings.Contains(hostname, "127.0.0.1") {
```
aliculPix4D, line 100: support tests...

cogito/ghcommitsink.go
```diff
@@ -93,9 +93,21 @@ func ghMakeContext(request PutRequest) string {
...
 		return fmt.Sprintf("http://%s", hostname)
 	}
-	return github.API
+	return fmt.Sprintf("https://%s/api/v3", hostname)
```
aliculPix4D, line 103: supports the GHEnterprise hostnames...

aliculPix4D: Personally, I find all the logic in: https://github.com/cli/cli/blob/trunk/internal/ghinstance/host.go#L24 
inverted for my taste..

cogito/ghcommitsink.go
```diff
@@ -93,9 +93,21 @@ func ghMakeContext(request PutRequest) string {
 }
 
 func apiEndpoint(h string) string {
```
aliculPix4D, line 95: maybe the better place would be to move these two functions in github pkg.

cogito/ghcommitsink.go
odormond: There must exist a go lib to just parse the URL and extract the hostname.

cogito/protocol.go
```diff
@@ -260,7 +260,9 @@ func (src *Source) Validate() error {
 	//
 	// Validate optional fields.
 	//
-	// In this case, nothing to validate.
+	if src.GhHostname != "" {
+		src.GhHostname = normalizeHostname(src.GhHostname)
```
odormond, line 264: Given your current implementation, you could call `normalizeHostname` in all cases as it would leave an empty string as empty string.

cogito/ghcommitsink.go
aliculPix4D: Hmm... yes it exists but I don't see how I can use it here. In `source.github_hostname` we expect the hostname it self, not URL e.g a string like `github.com`, `my-ghe.com` or `127.0.1:5435`. I want to add this normalizedHostname function (inspired by github cli tool code) to handle the cases when a user writes configures this variable in a way we didn't expect i.e by adding `/` or `/api/v1` to it...

Why return an error in that case to the user if we can easily recover from it?

odormond: You don't have to return an error because you use a tool that can return an error. Just handle it. :-)
Pseudo code (based on how python urlparse handle `"foo.bar.baz"`):
```
  parsed_url, err := url.parse(h)
  if err != nil {
    return h
  }
  if parsed_url.hostname != "" {
    return parse_url.hostname
  }
  if parsed_url.path != "" {
    return parsed_url.path.split("/")[0]
  }
```

aliculPix4D: My understanding is that you would need to inspect an error; so already your first check would be wrong.

We expect the user to configure: `github_hostname` by configuring `my-ghe.com` or test `127.0.1:1234` the url.parse would just return an error for both because there is no protocol configured. So in the first if statement I would need to cover inspect the error and continue if protocol is missing; and we don't even want the user to configure the protocol.

Note, if I expect the user to configure the protocol also so some kind of: `github_base_url` then the other part of the code needs to somehow strip the protocol. Additionally having `github_base_url` is useless because there is no easy way to construct the API endpoint for all 3 cases (default, test and enterprise).

README.md
```diff
@@ -198,6 +198,10 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
   The log level (one of `debug`, `info`, `warn`, `error`, `silent`).\
   Default: `info`.
 
+- `github_hostname`:\
+  Override the default GitHub hostname. This allows you to post commit statuses to repositories hosted by GitHub Enterprise (GHE) instances. e.g, `github.mycompany.org`, without protocol (only HTTPS is supported) and without API endpoint suffix (`/api/v3/`).
```
marco-m-pix4d, line 202: This allows you to post commit -> This allows to post commit

cogito/ghcommitsink.go
aliculPix4D: Discussed in other PR. I will move as suggested there.

README.md
marco-m-pix4d: > e.g,

Beginning of new phrase, needs uppercase. If it is a line break, it should be indented below "Override".

> Override the default GitHub hostname. 

There is no need to specify the concept of overriding, otherwise we should do the same on all settings with a default. We can simply say:

> GitHub hostname. This allows ...

> without API endpoint suffix (`/api/v3/`).

Calling it suffix is super confusing, since it is the prefix of the actual endpoint. We can say

> without API path prefix (`/api/v3/`).

Even better, I would keep it simpler:

> GitHub hostname. This allows to post commit statuses to repositories hosted by GitHub Enterprise (GHE) instances. For example: `github.mycompany.org` will be expanded by cogito to `https://github.mycompany.org/api/v3`.

cogito/ghcommitsink.go
marco-m-pix4d: The logic here is indeed direct. As long as it works, I agree it is simpler to understand, good!

cogito/protocol.go
marco-m-pix4d: If it is possible to always call normalizeHostname, it would be indeed better, because one less conditional to put in one's head to understand.

cogito/ghcommitsink.go
marco-m-pix4d: All this is a bit confusing and I agree with Olivier that _if_ we could use the Go stdlib function URL parse to detect errors it would be better.

To help understanding the intent, we can start from documenting the function. What is the intent of the function? Can we put in the documentation two examples input/output to help the reader?

README.md
aliculPix4D: Changed as suggested.

cogito/ghcommitsink.go
aliculPix4D: So in this iteration I removed this normalization function with two goals:
1. first to show you that its not really required by anyone. Documentation should be clear enough:
```
  GitHub hostname. This allows to post commit statuses to repositories hosted by GitHub Enterprise (GHE) instances. For example: github.mycompany.org will be expanded by cogito to https://github.mycompany.org/api/v3.
```
if you configure the `source.github_hostname` correctly (no protocol, no API prefix, no trailing slash) it will just work.
2. after the point 1 is clear. We can now discuss bringing the normalization function back in the PR if we agree that it brings value and I will try to write it in a better way.

My point of view was: lets help the user in case it misconfigures the cogito by not failing if the most common mistakes are done:
- protocol added to the hostname: `source.github_hostname = https://github.mycompany.org`
- the trailing slash: `source.github_hostname = https://github.mycompany.org/api/v3/`
- API prefix already added: `source.github_hostname = github.mycompany.org/api/v3/`
- or both protocol and API prefix: `source.github_hostname = https://github.mycompany.org/api/v3/`

If you see benefit in the this normalization function, I will try to bring it back.

Opinions?

github/commitstatus.go
```diff
@@ -197,8 +197,11 @@ func (cs CommitStatus) explainError(err error, state, sha, url string) error {
 // if hostname is overrriden in tests by localhost HTTP testserver it returns http://127.0.0.1
 func ApiRoot(h string) string {
```
aliculPix4D, line 198: changes here are enough to support GHE.

README.md
```diff
@@ -198,6 +198,9 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
   The log level (one of `debug`, `info`, `warn`, `error`, `silent`).\
   Default: `info`.
 
+- `github_hostname`:\
+  GitHub hostname. This allows to post commit statuses to repositories hosted by GitHub Enterprise (GHE) instances. For example: github.mycompany.org will be expanded by cogito to https://github.mycompany.org/api/v3.
```
aliculPix4D, line 202: add the default... removed by mistake..

github/commitstatus.go
```diff
@@ -197,8 +197,11 @@ func (cs CommitStatus) explainError(err error, state, sha, url string) error {
 // if hostname is overrriden in tests by localhost HTTP testserver it returns http://127.0.0.1
```
aliculPix4D, line 197: TODO:
- [ ] add the comment here for GHE

aliculPix4D:
@Pix4D/integration this PR is now ready for review.

SpectreVert (approved):

odormond (approved):

github/commitstatus.go
```diff
@@ -198,11 +198,14 @@ func (cs CommitStatus) explainError(err error, state, sha, url string) error {
 // Example:
 // if hostname is: github.com it returns https://api.github.com
 // if hostname is overrriden in tests by localhost HTTP testserver it returns http://127.0.0.1:PORT
```
marco-m-pix4d, line 200: // if hostname looks like a httptest server, it returns ...

github/commitstatus.go
```diff
@@ -198,11 +198,14 @@ func (cs CommitStatus) explainError(err error, state, sha, url string) error {
...
+// if hostname is Github Enterprise instance (e.g., github.mycompany.org) it returns e.g., https://github.mycompany.org/api/v3
```
marco-m-pix4d, line 201: // otherwise, hostname is assumed to be of a Github Enterprise instance. For example, github.mycompany.org returns https://github.mycompany.org/api/v3

marco-m-pix4d (approved):
🎉

github/commitstatus.go
aliculPix4D: Fixed.

github/commitstatus.go
```diff
@@ -198,11 +198,14 @@ func (cs CommitStatus) explainError(err error, state, sha, url string) error {
 // Example:
 // if hostname is: github.com it returns https://api.github.com
```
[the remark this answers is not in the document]
aliculPix4D: Fixed.

PR #149, opened 2024-02-18
fix: instantiated pipeline vars

### LINKED ISSUE #148: Build links for instanced pipelines don't work
We use [Instanced Pipelines](https://concourse-ci.org/instanced-pipelines.html), and the build links we get for them in GitHub don't appear to work.  Here is an example of a link from cogito:

`https://concourse/teams/teamname/pipelines/pipeline_name/jobs/job_name/builds/5?vars=%7B%22DEPLOYMENTID%22%3A%222%22%2C%22DESC%22%3A%22Daily+tests+for+Preprod+2%22%2C%22TITLE%22%3A%22PREPROD+TESTS%22%7D`

an example of a working link for the above build

`https://concourse/teams/teamname/pipelines/pipeline_name/jobs/job_name/builds/5?vars.DEPLOYMENTID=%222%22&vars.DESC=%22Daily%20tests%20for%20Preprod%202%22&vars.TITLE=%22PREPROD%20TESTS%22`

jamiegosling:
Pipeline variables for instantiated pipelines should be encoded as a sequence of URL parameters in the format like

`?vars.variableName1=variableValue1&vars.variableName2=variableValue2`

Closes #148

### DIFF
--- a/cogito/putter_private_test.go
+++ b/cogito/putter_private_test.go
@@ -345,13 +345,19 @@ func TestConcourseBuildURL(t *testing.T) {
 			name: "instanced vars 1",
 			env: testhelp.MergeStructs(baseEnv,
 				Environment{BuildPipelineInstanceVars: `{"branch":"stable"}`}),
-			want: "https://ci.example.com/teams/devs/pipelines/magritte/jobs/paint/builds/42?vars=%7B%22branch%22%3A%22stable%22%7D",
+			want: "https://ci.example.com/teams/devs/pipelines/magritte/jobs/paint/builds/42?vars.branch=%22stable%22",
 		},
 		{
 			name: "instanced vars 2",
 			env: testhelp.MergeStructs(baseEnv,
 				Environment{BuildPipelineInstanceVars: `{"branch":"stable","foo":"bar"}`}),
-			want: "https://ci.example.com/teams/devs/pipelines/magritte/jobs/paint/builds/42?vars=%7B%22branch%22%3A%22stable%22%2C%22foo%22%3A%22bar%22%7D",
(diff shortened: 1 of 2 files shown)


jamiegosling:
Thanks @marco-m-pix4d,

I don't really know Go so if I get the time I'll iterate here as a learning experience, but if you fix it before then then that's great.  Thanks for your time building this and looking at my issue!

Jamie

cogito/putter_private_test.go
```diff
@@ -345,13 +345,19 @@ func TestConcourseBuildURL(t *testing.T) {
 			name: "instanced vars 1",
 			env: testhelp.MergeStructs(baseEnv,
 				Environment{BuildPipelineInstanceVars: `{"branch":"stable"}`}),
-			want: "https://ci.example.com/teams/devs/pipelines/magritte/jobs/paint/builds/42?vars=%7B%22branch%22%3A%22stable%22%7D",
+			want: "https://ci.example.com/teams/devs/pipelines/magritte/jobs/paint/builds/42?vars.branch=%22stable%22",
```
aliculPix4D, line 348: @marco-m-pix4d should we start using go-vcr pkg in these tests?

I wonder if we could have catch this sooner if we used it already.

marco-m-pix4d: > should we start using go-vcr pkg in these tests?

Probably.

> I wonder if we could have catch'ed this sooner if we used it already.

This is the part I don't understand. I think that I tested against https://github.com/marco-m/concourse-in-a-box, so I don't understand why this now fails. True in any case that if we had a go-vcr trace, we could answer this question...

aliculPix4D:
@jamiegosling see: https://github.com/Pix4D/cogito/pull/153

aliculPix4D:
@jamiegosling I am closing this PR as this was fixed in: https://github.com/Pix4D/cogito/pull/153

Thanks for your contribution and reporting the issue.

PR #150, opened 2024-02-21
Update to Go 1.22 and update deps

marco-m-pix4d:


sets/sets.go
```diff
@@ -79,10 +79,3 @@ func (s *Set[T]) Difference(x *Set[T]) *Set[T] {
 	}
 	return result
 }
-
-func max(a, b int) int {
```
aliculPix4D, line 83: :)

aliculPix4D (approved):

sets/sets.go
marco-m-pix4d, line 83: This "simple" change needed a lot of work on the Go compiler...

SpectreVert: Why is that? I looked for articles about it but didn't find much.

SpectreVert (approved):

iAmoric (approved):

sets/sets.go
marco-m-pix4d: https://go.dev/ref/spec#Min_and_max

```
> The built-in functions min and max compute the smallest—or largest,
> respectively—value of a fixed number of arguments of ordered types
                                                       ^^^^^^^^^^^^^
```
https://go.dev/ref/spec#Comparison_operators defines "ordered types".

So the signature of the built-in max() is 
```
func max[T cmp.Ordered](x T, y ...T) T
```

and... `T` is a "type parameter", that is, Go "generics" ...

odormond (approved):

PR #152, opened 2024-03-14
replace github.com/hashicorp/go-hclog with stdlib log/slog

marco-m-pix4d:
"A promise is a promise", special for @aliculPix4D :-)


cmd/cogito/main.go
```diff
@@ -37,16 +36,27 @@ func mainErr(in io.Reader, out io.Writer, logOut io.Writer, args []string) error
...
 	}
-	log := hclog.New(&hclog.LoggerOptions{
-		Name:        "cogito",
-		Level:       hclog.LevelFromString(logLevel),
-		Output:      logOut,
-		DisableTime: true,
-	})
+	var level slog.Level
+	if err := level.UnmarshalText([]byte(logLevel)); err != nil {
+		return fmt.Errorf("%s. (valid: DEBUG, INFO, WARN, ERROR)", err)
```
marco-m-pix4d, line 46: Ah.
I think I need to double-check this :-(

aliculPix4D:
> "A promise is a promise", special for @aliculPix4D :-)

Thank you :)

cmd/cogito/main.go
aliculPix4D: Why?

You did the same in skyguide. It works fine.
https://github.com/Pix4D/skyguide/blob/28ac17a5f0409d622fec7ee1bd92b24dfdb76add/pkg/skyguide/app.go#L69

cmd/cogito/main.go
```diff
@@ -37,16 +36,27 @@ func mainErr(in io.Reader, out io.Writer, logOut io.Writer, args []string) error
...
+	}
+	removeTime := func(groups []string, a slog.Attr) slog.Attr {
```
aliculPix4D, line 48: Just for my understanding: why you don't want the timestamp?

aliculPix4D (approved):

cmd/cogito/main.go
marco-m-pix4d: We are disabling timestamps in two places: in prod code (here) and in the tests.

In the tests, because timestamps make tests more difficult, when doing string matching.

In prod code (here), also with hclog we were disabling timestamps, because they are already provided by the Concourse UI. See the removed line 48 just above.

I don't have a strong opinion, if we think it would help adding them in prod (this would allow for example to have them when doing a mouse copy and paste)...

cmd/cogito/main.go
marco-m-pix4d: I had a vague recollection we were doing something particular with the log levels. I now reconstructed the situation, see the second commit I pushed.

Do not be confused by the fact that I also made the strings lowercase, that is purely cosmetic to better match with the README, both cases are accepted. The real difference is not in the code but in the changes to the README and CHANGELOG.

aliculPix4D: I see. Thanks. Good catch.

cmd/cogito/main.go
aliculPix4D: Ok. Lets keep the same behavior. It will be simple to add later the timestamp if we see the need. 
Thanks for the explanation.

SpectreVert (approved):

iAmoric (approved):

PR #153, opened 2024-03-22
cogito: properly encode url

aliculPix4D:
cogito: properly encode url

Actually, the only problem that existed was due the way spaces were encoded. Even the most complex case with nested json worked fine with the existing code. The only thing "missing" was the proper encoding of spaces.

Concourse requires that the space characters are encoded as %20 instead of +. On the other hand golangs url.QueryEscape as well as url.Values.Encode() both encode the space as a + character.
See: https://github.com/golang/go/issues/4013

Fixes: https://github.com/Pix4D/cogito/issues/148

### DIFF
--- a/cogito/putter.go
+++ b/cogito/putter.go
@@ -333,6 +333,12 @@ func concourseBuildURL(env Environment) string {
 	// https://ci.example.com/teams/main/pipelines/cogito/jobs/autocat/builds/3?vars=%7B%22branch%22%3A%22stable%22%7D
 	if env.BuildPipelineInstanceVars != "" {
 		buildURL += fmt.Sprintf("?vars=%s", url.QueryEscape(env.BuildPipelineInstanceVars))
+		// Concourse requires that the space characters
+		// are encoded as %20 instead of +. On the other hand
+		// golangs url.QueryEscape as well as url.Values.Encode()
+		// both encode the space as a + character.
+		// See: https://github.com/golang/go/issues/4013
+		buildURL = strings.ReplaceAll(buildURL, "+", "%20")
 	}
 
 	return buildURL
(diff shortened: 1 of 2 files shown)


aliculPix4D:
See:
Screenshot from 2024-03-22 10-55-58

the link for autocat (from my local concourse deployment) correctly points to the build. The pipeline was worst case scenario possible: nested json with spaces

Sadly, we can't test this better in our Concourse deployment because we don't have `CONCOURSE_ENABLE_PIPELINE_INSTANCES` enabled.

cogito/putter.go
```diff
@@ -333,6 +333,12 @@ func concourseBuildURL(env Environment) string {
 	// https://ci.example.com/teams/main/pipelines/cogito/jobs/autocat/builds/3?vars=%7B%22branch%22%3A%22stable%22%7D
 	if env.BuildPipelineInstanceVars != "" {
 		buildURL += fmt.Sprintf("?vars=%s", url.QueryEscape(env.BuildPipelineInstanceVars))
+		// Concourse requires that the space characters
+		// are encoded as %20 instead of +. On the other hand
+		// golangs url.QueryEscape as well as url.Values.Encode()
+		// both encode the space as a + character.
+		// See: https://github.com/golang/go/issues/4013
+		buildURL = strings.ReplaceAll(buildURL, "+", "%20")
```
aliculPix4D, line 341: If we have a + anywhere else e.g in branch name: "sta+ble" the + would first be escaped with %2B and only remaining + in URL would be the one to escape the space. So this should be safe enough.

odormond (approved):

cogito/putter.go
marco-m: Sigh. I followed all the links from that issue and I have to agree there isn't any hope. It is also beyond my understanding how Concourse, written in Go, does not accept `+` in the query parameters part.

Thanks for having attempted to do something better.

marco-m (approved):

marco-m-pix4d (approved):

PR #155, opened 2024-05-13
sets: add sets intersection

marco-m-pix4d:
Part of PCI-3790

### DIFF
--- a/sets/sets.go
+++ b/sets/sets.go
@@ -84,17 +84,14 @@ func (s *Set[T]) Difference(x *Set[T]) *Set[T] {
 func (s *Set[T]) Intersection(x *Set[T]) *Set[T] {
 	result := New[T](0)
 	// loop over the smaller set (thanks to https://github.com/deckarep/golang-set)
-	if s.Size() < x.Size() {
-		for item := range s.items {
-			if x.Contains(item) {
-				result.items[item] = struct{}{}
-			}
-		}
-	} else {
-		for item := range x.items {
-			if s.Contains(item) {
-				result.items[item] = struct{}{}
-			}
+	smaller := s
+	bigger := x
+	if smaller.Size() > bigger.Size() {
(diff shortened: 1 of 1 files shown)


sets/sets.go
```diff
@@ -72,9 +72,29 @@ func (s *Set[T]) Remove(item T) bool {
...
+				result.items[item] = struct{}{}
+			}
+		}
+	} else {
+		for item := range x.items {
+			if s.Contains(item) {
+				result.items[item] = struct{}{}
+			}
 		}
 	}
```
odormond, line 99: Wouldn't it be better to just swap the variables and code the loop only once?
```suggestion
	if s.Size() > x.Size() {
		// Swap them to loop over the smallest set
		s, x = x, s
	}
	for item := range s.items {
		if x.Contains(item) {
			result.items[item] = struct{}{}
		}
	}
```

odormond (approved):

aliculPix4D (approved):
I remeber doing these exercises from Exercism :)

SpectreVert (approved):

iAmoric (approved):
Just for my understanding, what is the relation between this and [PCI-3790](https://pix4dbug.atlassian.net/browse/PCI-3790) ?

sets/sets.go
marco-m-pix4d: This is swapping pointers (`s` is a pointer):
```
func (s *Set[T]) Intersection(...)
```

 but on the other hand, in Go, "assignment to the method receiver propagates only to callees but not to callers", so at the exit of the function, the contents of the set do remain the same. Said in another way, the code proposed is safe, although confusing.

marco-m-pix4d: Just wanted to notice that GitHub itself decided that "marco-m-pix4d marked this conversation resolved".
I did nothing 🙄.

I added a second commit, that I thinks makes the code less confusing.

PR #158, opened 2024-09-25
sets: add Union()

marco-m-pix4d:
This is needed for the facchino tags preview.

PCI-3992

### DIFF
--- a/sets/sets.go
+++ b/sets/sets.go
@@ -105,3 +105,15 @@ func (s *Set[T]) Intersection(x *Set[T]) *Set[T] {
 	}
 	return result
 }
+
+// Union returns a set containing all the elements of s and x.
+func (s *Set[T]) Union(x *Set[T]) *Set[T] {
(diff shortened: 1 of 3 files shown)


aliculPix4D (approved):

sets/sets.go
```diff
@@ -105,3 +105,15 @@ func (s *Set[T]) Intersection(x *Set[T]) *Set[T] {
 	}
 	return result
 }
+
+// Union returns a set containing all the elements of s and x.
+func (s *Set[T]) Union(x *Set[T]) *Set[T] {
```
aliculPix4D, line 110: Don't you also need the set difference as you wrote in chat? or we already have that...

aliculPix4D: Ah we have it... 
`func (s *Set[T]) Difference(x *Set[T]) *Set[T]`

iAmoric (approved):

PR #160, opened 2024-12-06
Add optional omit_target_url

### LINKED ISSUE #159: Feature Request: Optional target_url
I would love to be able to **optionally** set the target_url when setting the GH status. One of the Concourse CI systems I run is entirely private, but it does run pipelines for a public repository on GitHub. I'd like to be able to report build status, but without the URL as that is private information. Could define the optional target URL either in the resource configuration or on the `put`.

```yaml
    on_success:
      put: gh-status
      inputs: [the-repo]
      params: {state: success, target_url: false}

---

- name: gh-status
  type: cogito
  check_every: never
  source:
    owner: ((github-owner))
    repo: ((your-repo-name))
    access_token: ((github-PAT))
    target_url: false
```

marco-m-pix4d:
While at it, did some cleanups and updated to Go 1.23

Closes #159.

### DIFF
--- a/cogito/ghcommitsink.go
+++ b/cogito/ghcommitsink.go
@@ -54,6 +54,9 @@ func (sink GitHubCommitStatusSink) Send() error {
 		"state", ghState, "owner", sink.Request.Source.Owner,
 		"repo", sink.Request.Source.Repo, "git-ref", sink.GitRef,
 		"context", context, "buildURL", buildURL, "description", description)
+	if sink.Request.Source.OmitTargetURL {
+		buildURL = ""
+	}
 	if err := commitStatus.Add(sink.GitRef, ghState, buildURL, description); err != nil {
 		return err
 	}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,13 +13,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Install Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
         with:
           go-version: ${{ env.go-version }}
       - name: Install Task
         run: go install github.com/go-task/task/v3/cmd/task@${{ env.task-version }}
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           # By default, actions/checkout will persist the GITHUB_TOKEN, so that further
           # steps in the job can perform authenticated git commands (that is: WRITE to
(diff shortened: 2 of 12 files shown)


cogito/ghcommitsink.go
```diff
@@ -54,6 +54,9 @@ func (sink GitHubCommitStatusSink) Send() error {
 		"state", ghState, "owner", sink.Request.Source.Owner,
 		"repo", sink.Request.Source.Repo, "git-ref", sink.GitRef,
 		"context", context, "buildURL", buildURL, "description", description)
+	if sink.Request.Source.OmitTargetURL {
+		buildURL = ""
```
odormond, line 58: As the GH API expect a string or null, wouldn't be cleaner to pass null instead?
(https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#create-a-commit-status)

odormond: I'm also tempted to do this before logging to avoid a misleading `buildURL` showing up in the logs

odormond (approved):

aliculPix4D (approved):

SpectreVert (approved):

cogito/ghcommitsink.go
marco-m-pix4d: > I'm also tempted to do this before logging to avoid a misleading `buildURL` showing up in the logs

That was my first iteration. Then I realized that the logging should still contain the buildURL, since the logging is not publicly accessible, and the goal of this PR is to omit the Concourse URL from a public GitHub PR page. Omitting also from the log would, in my understanding, make troubleshooting or auditing more difficult for no reason...

marco-m-pix4d: > As the GH API expect a string or null, wouldn't it be cleaner to pass null instead? (https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#create-a-commit-status)

The type of buildURL is a string, not a pointer to a string, so we cannot set it to nil. We could change the type to pointer to string, but that would bring pain everywhere else.

What we could do is to decorate the structure field `TargetURL` with tag `omitempty`

https://github.com/Pix4D/cogito/blob/a55a8121dbcc3b9ad96c573d0a37dcfc93bdf8f7/github/commitstatus.go#L83-L89

like this:

```
	TargetURL   string `json:"target_url,omitempty"`
```

`omitempty` will see an empty string (the value `""`) and will omit the field from the JSON object. This is not the same as setting the field to JSON null...

To summarize, I am tempted to say that if the GH API is happy with an empty string, that is good enough. But this discussion made me realize that I need to test this somehow.

marco-m-pix4d: @odormond I am mentioning you here because I am going to merge this PR. I believe I addressed both your comments. If I missed something, ping me.

Issue #151, opened 2024-02-28
Support GH App Installation Access


tgoodsell-tempus:
GH App Installations have a much larger API quota compared with user PATs.

Would be nice if this supported natively authing as a GH app installation and using that access token.

Docs: https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation

marco-m-pix4d:
@tgoodsell-tempus Ah, I never though about that!
I am not very familiar with GH Apps. My first reaction is: would that work also considering that cogito is very transient (container lifetime is < 10 sec) and many instances can exist at the same time in a Concourse installation?
Or am I misunderstanding your idea?

tgoodsell-tempus:
@marco-m-pix4d Yes, provided you treat it similar to the PAT where you just have the end user provide all of the:
- App ID
- Installation ID
- App Private Key
- Org ID

A really basic code example for setting up the GH golang client to use this would be:
```go
tr := http.DefaultTransport
transport, err := ghinstallation.New(tr, "USER-APP-ID", "USER-INSTALL-ID", []byte("USER-APP-PRIVATE-KEY"))
if err != nil {
	return nil, org, err
}
client := github.NewClient(&http.Client{Transport: transport})
```

Just treat it as an alternative to the PAT. I don't think this warrants you doing something like creating a "central" app something like a managed service that runs on a client's Github account uses to manage itself.

marco-m-pix4d:
> Just treat it as an alternative to the PAT. 

Ah, I think this is the beef. Good idea, thanks! We will investigate.

scottillogical:
> EDIT: to check also what happens in case of concurrency. Currenly, we use a user PAT, and concurrency works fine. Since this is a GH "app", I wonder if GH side would do any check. This is easy to validate before going all-in in the implementation.

In our testing concurrency effectively works better with github app vs PAT because the rate limit is higher

aliculPix4D:
Will be released in: https://github.com/Pix4D/cogito/pull/167

PR #161, opened 2025-04-02
add github app auth

scottillogical:
TODO:
- Add taskfile support
- For CI to work, create a github app and add secrets to CI secrets backend


marco-m-pix4d:
Hello @scottillogical, thanks for your contribution.

Quoting from the README:

> Please, before opening a PR, open a ticket to discuss your use case. This allows to better understand the why of a new feature and not to waste your time (and ours) developing a feature that for some reason doesn't fit well with the spirit of the project or could be implemented differently. This is in the spirit of [Talk, then code](https://dave.cheney.net/2019/02/18/talk-then-code).

Is this PR related to #151 ?

Thanks!

scottillogical:
> Hello @scottillogical, thanks for your contribution.
> 
> Quoting from the README:
> 
> > Please, before opening a PR, open a ticket to discuss your use case. This allows to better understand the why of a new feature and not to waste your time (and ours) developing a feature that for some reason doesn't fit well with the spirit of the project or could be implemented differently. This is in the spirit of [Talk, then code](https://dave.cheney.net/2019/02/18/talk-then-code).
> 
> Is this PR related to #151 ?
> 
> Thanks!

Yes it is sorry I intended to reference that issue directly!

README.md
```diff
@@ -143,7 +143,7 @@ The absolute minimum is setting keys `gchat_webhook` and `sinks` in the `source`
 
 ## Build states mapping
 
-There is only a partial matching between Concourse and GitHub Commit status API, so we map as good as we can according to this table:
```
scottillogical, line 146: Trailing whitespace

marco-m-pix4d:
@scottillogical we are a bit under pressure so reviewing will take some time, but we want to come back to you. Please feel free to ping here in a week if no activity. Thanks.

github/commitstatus.go
```diff
@@ -69,14 +74,18 @@ type CommitStatus struct {
 //
 // See also:
 // - https://docs.github.com/en/rest/commits/statuses
-func NewCommitStatus(target *Target, token, owner, repo, context string, log *slog.Logger) CommitStatus {
+func NewCommitStatus(target *Target, token, owner, repo, context string, log *slog.Logger, useGithubAppAuth bool, privateKey string, applicationId, installationId int64) CommitStatus {
```
aliculPix4D, line 77: I would like to avoid having so many function arguments...

github/commitstatus.go
```diff
@@ -121,14 +130,29 @@ func (cs CommitStatus) Add(sha, state, targetURL, description string) error {
...
+		req.Header.Set("Authorization", "token "+cs.token)
+	}
 	req.Header.Set("Accept", "application/vnd.github.v3+json")
 	req.Header.Set("Content-Type", "application/json")
 
 	// The retryable unit of work.
 	workFn := func() error {
 		// By default, there is no timeout, so the call could hang forever.
 		client := &http.Client{Timeout: time.Second * 30}
+		if cs.useGithubAppAuth {
```
aliculPix4D, line 143: In my opinion, this is the wrong place to do this... (not the fault of this PR)

Our initial implementation was wrong and we should not initialize the client so low in the chain...

github/commitstatus.go
scottillogical: i thought about that too, we could have NewCommitStatusGithubApp and NewCommitStatusAccessToken i suppose? other than that, I am not sure of how to avoid the issue

github/commitstatus.go
```diff
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"github.com/Pix4D/cogito/retry"
+	"github.com/bradleyfalzon/ghinstallation"
```
aliculPix4D, line 21: seems small enough... but we need to check if its really crucial to bring in this dependency...

github/commitstatus.go
aliculPix4D: I have some ideas, but need to first understand in detail what `github.com/bradleyfalzon/ghinstallation` does...

aliculPix4D (commented):
@scottillogical Nice work. Thanks for adding this feature.

I just had a quick look at the PR and left some comments, more for myself than something to address. 

I am off until Tuesday. But when back I plan to check if it would be worthwhile to make a refactoring PR that would make implementing this feature even more straightforward.

CONTRIBUTING.md
```diff
@@ -81,6 +107,28 @@ The integration tests (tests that interact with GitHub) are disabled by default
...
+### Add github app auth credentials
+
+You can either generate an app for your individual account, or create a GitHub organization and assign that as the application owner
+
+1. Create a test repo
+https://github.com/new
+2. Create a new github app
+https://github.com/settings/apps/new
+* Mark webhook as inactive)
+* Add repoistory permission: Commit Statuses: Read/write
```
aliculPix4D, line 119: typo: repoistory

README.md
```diff
@@ -183,10 +183,32 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
...
 - `access_token`\
   The OAuth access token.\
   See also: section [GitHub OAuth token](#github-oauth-token).
 
+#### GitHub app keys
+
+- `private_key` \
+  The private key generated for the GitHub application
+
+- `application_id`
```
aliculPix4D, line 204: According to the GitHub docs... this can be app id or client id... client id is preffered.

README.md
```diff
@@ -183,10 +183,32 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
...
+  The application id of the GitHub application
+
+- `installation_id`
```
aliculPix4D, line 207: How do you find the installation id? The only way I managed to achieve this is with my custom script that:
- creates a signed JWT token
- do a GET request using the JWT to: `/app/installations` endpoint

The docs here:
https://github.com/bradleyfalzon/ghinstallation?tab=readme-ov-file#what-is-app-id-and-installation-id

is wrong...

> Settings > Developer > settings > GitHub Apps > Advanced > Payload in Request tab

doesn't exists.. at least not for me with my test app and webhooks inactive... (which is what you are suggesting a few lines above..)

scottillogical: If you go to 

https://github.com/settings/apps/test-cogito/installations
and then click the gear icon  its in the URL
https://github.com/settings/installations/<ID>

README.md
scottillogical: client id is not yet supported by https://github.com/bradleyfalzon/ghinstallation/issues/118

README.md
aliculPix4D: I don't have access to these pages... Can you post a screenshot?

Just curious:
- is this organization app or your user app?
- are webhooks active on it or not?

When I visit those pages for my test app, there is no mention of installation ids... only client id and app id...

scottillogical: i tested with both and organization app and user app.   webhooks are not required.    installation id is in the URL that the configure button takes you too

scottillogical: i..e https://github.com/settings/installations/63671192 the ID is the installation id

aliculPix4D: Yes... it really is :) it was not evident for me, I must say. Thanks for clarifying this..

We will need to update the docs explaining in detail (even with screenshot) if necessary how to find this value. We should probably also offer the alternative way with a one-off script...

github/commitstatus.go
```diff
@@ -121,14 +130,29 @@ func (cs CommitStatus) Add(sha, state, targetURL, description string) error {
...
 	req.Header.Set("Content-Type", "application/json")
 
 	// The retryable unit of work.
 	workFn := func() error {
 		// By default, there is no timeout, so the call could hang forever.
 		client := &http.Client{Timeout: time.Second * 30}
+		if cs.useGithubAppAuth {
+			// We need a transport that we can update if using GitHub App authentication
+			transport := http.DefaultTransport.(*http.Transport).Clone()
+			var ghAppInstallationTransport *ghinstallation.Transport
```
aliculPix4D, line 146: NOTE: if I am not mistaken, as it is implemented here, this would be breaking for our GitHub Enterprise users since ghinstallation would always try to generate installation token against `https://api.github.com`

cogito/protocol.go
```diff
@@ -164,8 +164,17 @@ type Source struct {
...
+	// Mandatory if not using github app auth
 	AccessToken string `json:"access_token"` // SENSITIVE
-	//
+
+	//Mandtory if using github app auth
+
+	UseGithubAppAuth bool `json:"use_github_app_auth"`
+	ApplicationId int64 `json:"application_id"`
+	InstallationId int64 `json:"installation_id"`
+	PrivateKey string `json:"private_key"` // SENSITIVE
```
aliculPix4D, line 176: NOTE: missing redact on this field... As it is it would leak private key in logs... especially in debug mode...

aliculPix4D:
Just a status update: I am working on another implementation of this feature... It is still work in progress, and definitely not ready yet, but you can follow the work in:
https://github.com/Pix4D/cogito/compare/master...pci-4263-support-github-apps

The main difference compared to your approach: 
- we are not using the `bradleyfalzon/ghinstallation` package but instead a function to generate the installation token is written from scratch...  This ensures that we don't play with http transports and round trippers... and it fits more naturally to the existing cogito code... since we just have to generate token and pass it around in the same way we were passing the access token before...

aliculPix4D:
@scottillogical I think we can now close this PR...

Thank you very much for your contribution... It really helped us to kickstart and implement this feature more quickly.

FYI I plan to release v0.13.0 today or at latest on Monday (https://github.com/Pix4D/cogito/pull/167)...

aliculPix4D:
@scottillogical  v0.13.0 is out... so I am closing this PR. Thanks again for the PR and providing us with the original implementation for this feature. Sorry for the wait...

PR #162, opened 2025-04-24
github: add support for github apps

aliculPix4D:
GitHub: add support for GitHub apps

Docker image available for testing: `pix4d/cogito:pci-4263-support-github-apps`

Passes all integration and acceptance tests in CI pipeline...

It's still a draft because there are still some things to polish like:
- ~~update docs about github app, increased rate limit and how to configure it~~
- ~~private_key: multiline or base64 encoded? so far, seems fine with multiline...~~
- ~~print in logs rate limit after each request (at least in debug)~~
- ~~update acceptance tests pipelines according to our OPA rules~~
- ~~update pipeline with acceptance tests for github app~~
- maybe implement retry on a POST HTTP request that fetches the installation token
etc...

Some stuff from the list above are optional, some required, like updating docs, but since this PR is fully functional already, I would like to start collecting review feedback.

Basically, it replaces: https://github.com/Pix4D/cogito/pull/161 and closes: https://github.com/Pix4D/cogito/issues/151

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -57,6 +57,10 @@ tasks:
       COGITO_TEST_REPO_OWNER: '{{default "pix4d" .COGITO_TEST_REPO_OWNER}}'
       COGITO_TEST_GCHAT_HOOK:
         sh: 'echo {{default "$(gopass show cogito/test_gchat_webhook)" .COGITO_TEST_GCHAT_HOOK}}'
+      COGITO_TEST_GH_APP_CLIENT_ID: '{{default "Iv23lir9pyQlqmweDPbz" .COGITO_TEST_GH_APP_CLIENT_ID}}'
+      COGITO_TEST_GH_APP_INSTALLATION_ID: '{{default "64650729" .COGITO_TEST_GH_APP_INSTALLATION_ID}}'
+      COGITO_TEST_GH_APP_PRIVATE_KEY:
+        sh: 'echo "{{default "$(gopass show cogito/test_gh_app_private_key)" .COGITO_TEST_GH_APP_PRIVATE_KEY}}"'
 
   test:unit:
     desc: Run the unit tests.
--- a/cmd/cogito/main_test.go
+++ b/cmd/cogito/main_test.go
@@ -141,7 +141,7 @@ func TestRunPutGhAppSuccessIntegration(t *testing.T) {
 	}
 
 	gitHubCfg := testhelp.GitHubSecretsOrFail(t)
-	ghAppInstallationID, err := strconv.ParseInt(gitHubCfg.GhAppInstallationID, 10, 64)
+	ghAppInstallationID, err := strconv.ParseInt(gitHubCfg.GhAppInstallationID, 10, 32)
 	assert.NilError(t, err)
 
 	stdin := bytes.NewReader(testhelp.ToJSON(t, cogito.PutRequest{
@@ -150,7 +150,7 @@ func TestRunPutGhAppSuccessIntegration(t *testing.T) {
 			Repo:  gitHubCfg.Repo,
 			GitHubApp: github.GitHubApp{
 				ClientId:       gitHubCfg.GhAppClientID,
-				InstallationId: ghAppInstallationID,
+				InstallationId: int(ghAppInstallationID),
 				PrivateKey:     gitHubCfg.GhAppPrivateKey,
 			},
 			LogLevel: "debug",
--- a/cmd/cogito/main_test.go
+++ b/cmd/cogito/main_test.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strconv"
 	"strings"
 	"testing"
 	"testing/iotest"
@@ -132,6 +133,46 @@ func TestRunPutSuccessIntegration(t *testing.T) {
 		`level=INFO msg="commit status posted successfully" name=cogito.put name=ghCommitStatus state=error`))
 	assert.Assert(t, cmp.Contains(stderr.String(),
 		`level=INFO msg="state posted successfully to chat" name=cogito.put name=gChat state=error`))
+}
+
+func TestRunPutGhAppSuccessIntegration(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration test (reason: -short)")
+	}
+
+	gitHubCfg := testhelp.GitHubSecretsOrFail(t)
+	ghAppInstallationID, err := strconv.ParseInt(gitHubCfg.GhAppInstallationID, 10, 64)
+	assert.NilError(t, err)
+
+	stdin := bytes.NewReader(testhelp.ToJSON(t, cogito.PutRequest{
+		Source: cogito.Source{
+			Owner: gitHubCfg.Owner,
+			Repo:  gitHubCfg.Repo,
+			GitHubApp: github.GitHubApp{
+				ClientId:       gitHubCfg.GhAppClientID,
+				InstallationId: ghAppInstallationID,
+				PrivateKey:     gitHubCfg.GhAppPrivateKey,
+			},
+			LogLevel: "debug",
+		},
+		Params: cogito.PutParams{State: cogito.StateError},
+	}))
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+	inputDir := testhelp.MakeGitRepoFromTestdata(t, "../../cogito/testdata/one-repo/a-repo",
+		testhelp.HttpsRemote(github.GhDefaultHostname, gitHubCfg.Owner, gitHubCfg.Repo), gitHubCfg.SHA,
+		"ref: refs/heads/a-branch-FIXME")
+	t.Setenv("BUILD_JOB_NAME", "TestRunPutGhAppSuccessIntegration")
+	t.Setenv("ATC_EXTERNAL_URL", "https://cogito.invalid")
+	t.Setenv("BUILD_PIPELINE_NAME", "the-test-pipeline")
+	t.Setenv("BUILD_TEAM_NAME", "the-test-team")
+	t.Setenv("BUILD_NAME", "42")
+
+	err = mainErr(stdin, &stdout, &stderr, []string{"out", inputDir})
+
+	assert.NilError(t, err, "\nstdout:\n%s\nstderr:\n%s", stdout.String(), stderr.String())
+	assert.Assert(t, cmp.Contains(stderr.String(),
+		`level=INFO msg="commit status posted successfully" name=cogito.put name=ghCommitStatus state=error`))
 }
 
 func TestRunFailure(t *testing.T) {
--- a/cogito/check_test.go
+++ b/cogito/check_test.go
@@ -8,6 +8,7 @@ import (
 	"gotest.tools/v3/assert"
 
 	"github.com/Pix4D/cogito/cogito"
+	"github.com/Pix4D/cogito/github"
 	"github.com/Pix4D/cogito/testhelp"
 )
 
@@ -88,10 +89,25 @@ func TestCheckFailure(t *testing.T) {
 
 	testCases := []testCase{
 		{
-			name:    "validation failure: missing repo keys",
+			name:    "validation failure: missing access token or github_app",
 			source:  cogito.Source{},
 			writer:  io.Discard,
-			wantErr: "check: source: missing keys: owner, repo, access_token",
+			wantErr: "check: source: one of access_token or github_app must be specified",
+		},
+		{
+			name: "validation failure: both access_key and github_app are set",
+			source: cogito.Source{
+				AccessToken: "dummy-token",
+				GitHubApp:   github.GitHubApp{ClientId: "client-id"},
+			},
+			writer:  io.Discard,
+			wantErr: "check: source: cannot specify both github_app and access_token",
+		},
+		{
+			name:    "validation failure: missing repo keys",
+			source:  cogito.Source{AccessToken: "dummy-token"},
+			writer:  io.Discard,
+			wantErr: "check: source: missing keys: owner, repo",
 		},
 		{
 			name: "validation failure: missing gchat keys",
--- a/cogito/ghcommitsink.go
+++ b/cogito/ghcommitsink.go
@@ -46,8 +46,10 @@ func (sink GitHubCommitStatusSink) Send() error {
 	server := github.ApiRoot(sink.Request.Source.GhHostname)
 
 	token := sink.Request.Source.AccessToken
+	// if access token is not configured, we are using github_app
+	// so we must generate the installation token
 	if token == "" {
-		installationToken, err := github.GenerateInstallationToken(server, sink.Request.Source.GitHubApp)
+		installationToken, err := github.GenerateInstallationToken(ctx, httpClient, server, sink.Request.Source.GitHubApp)
 		if err != nil {
 			return err
 		}
--- a/cogito/ghcommitsink.go
+++ b/cogito/ghcommitsink.go
@@ -43,10 +43,20 @@ func (sink GitHubCommitStatusSink) Send() error {
 	ghState := ghAdaptState(sink.Request.Params.State)
 	buildURL := concourseBuildURL(sink.Request.Env)
 	context := ghMakeContext(sink.Request)
+	server := github.ApiRoot(sink.Request.Source.GhHostname)
+
+	token := sink.Request.Source.AccessToken
+	if token == "" {
+		installationToken, err := github.GenerateInstallationToken(server, sink.Request.Source.GitHubApp)
+		if err != nil {
+			return err
+		}
+		token = installationToken
+	}
 
 	target := &github.Target{
 		Client: httpClient,
-		Server: github.ApiRoot(sink.Request.Source.GhHostname),
+		Server: server,
 		Retry: retry.Retry{
 			FirstDelay:   retryFirstDelay,
 			BackoffLimit: retryBackoffLimit,
@@ -54,7 +64,7 @@ func (sink GitHubCommitStatusSink) Send() error {
 			Log:          sink.Log,
 		},
 	}
-	commitStatus := github.NewCommitStatus(target, sink.Request.Source.AccessToken,
+	commitStatus := github.NewCommitStatus(target, token,
 		sink.Request.Source.Owner, sink.Request.Source.Repo, context, sink.Log)
 	description := "Build " + sink.Request.Env.BuildName
 
--- a/cogito/ghcommitsink_test.go
+++ b/cogito/ghcommitsink_test.go
@@ -1,6 +1,8 @@
 package cogito_test
 
 import (
+	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
@@ -28,7 +30,7 @@ func TestSinkGitHubCommitStatusSendSuccess(t *testing.T) {
 		Log:    testhelp.MakeTestLog(),
 		GitRef: wantGitRef,
 		Request: cogito.PutRequest{
-			Source: cogito.Source{GhHostname: gitHubSpyURL.Host},
+			Source: cogito.Source{GhHostname: gitHubSpyURL.Host, AccessToken: "dummy-token"},
 			Params: cogito.PutParams{State: wantState},
 			Env:    cogito.Environment{BuildJobName: jobName},
 		},
@@ -39,6 +41,62 @@ func TestSinkGitHubCommitStatusSendSuccess(t *testing.T) {
 	assert.NilError(t, err)
 	ts.Close() // Avoid races before the following asserts.
 	assert.Equal(t, path.Base(URL.Path), wantGitRef)
+	assert.Equal(t, ghReq.State, string(wantState))
+	assert.Equal(t, ghReq.Context, wantContext)
+}
+
+func TestSinkGitHubCommitStatusSendGhAppSuccess(t *testing.T) {
+	wantGitRef := "deadbeefdeadbeef"
+	wantState := cogito.StatePending
+	jobName := "the-job"
+	wantContext := jobName
+	var ghReq github.AddRequest
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.String() == "/repos/statuses/deadbeefdeadbeef" {
+			dec := json.NewDecoder(r.Body)
+			if err := dec.Decode(&ghReq); err != nil {
+				w.WriteHeader(http.StatusTeapot)
+				fmt.Fprintln(w, "test: decoding request:", err)
+				return
+			}
+		}
+
+		w.WriteHeader(http.StatusCreated)
+		if r.URL.String() == "/app/installations/12345/access_tokens" {
+			fmt.Fprintln(w, `{"token": "dummy_installation_token"}`)
+			return
+		}
+	}
+	ts := httptest.NewServer(http.HandlerFunc(handler))
+	defer ts.Close()
+
+	gitHubSpyURL, err := url.Parse(ts.URL)
+	assert.NilError(t, err, "error parsing SpyHttpServer URL: %s", err)
+
+	privateKey, err := testhelp.GeneratePrivateKey(t, 2048)
+	assert.NilError(t, err)
+
+	sink := cogito.GitHubCommitStatusSink{
+		Log:    testhelp.MakeTestLog(),
+		GitRef: wantGitRef,
+		Request: cogito.PutRequest{
+			Source: cogito.Source{
+				GhHostname: gitHubSpyURL.Host,
+				GitHubApp: github.GitHubApp{
+					ClientId:       "client-id",
+					InstallationId: 12345,
+					PrivateKey:     string(testhelp.EncodePrivateKeyToPEM(privateKey)),
+				},
+			},
+			Params: cogito.PutParams{State: wantState},
+			Env:    cogito.Environment{BuildJobName: jobName},
+		},
+	}
+
+	err = sink.Send()
+
+	assert.NilError(t, err)
 	assert.Equal(t, ghReq.State, string(wantState))
 	assert.Equal(t, ghReq.Context, wantContext)
 }
@@ -55,7 +113,7 @@ func TestSinkGitHubCommitStatusSendFailure(t *testing.T) {
 		Log:    testhelp.MakeTestLog(),
 		GitRef: "deadbeefdeadbeef",
 		Request: cogito.PutRequest{
-			Source: cogito.Source{GhHostname: gitHubSpyURL.Host},
+			Source: cogito.Source{GhHostname: gitHubSpyURL.Host, AccessToken: "dummy-token"},
 			Params: cogito.PutParams{State: cogito.StatePending},
 		},
 	}
--- a/cogito/protocol.go
+++ b/cogito/protocol.go
@@ -168,7 +168,7 @@ type Source struct {
 	AccessToken string `json:"access_token"` // SENSITIVE
 
 	// Mandatory if using github app auth
-	GitHubApp github.GitHubApp `json:"github_app,omitempty"`
+	GitHubApp github.GitHubApp `json:"github_app"`
 
 	//
 	// Optional
@@ -247,6 +247,15 @@ func (src *Source) Validate() error {
 	sinks := sets.From(src.Sinks...)
 	if sinks.Size() == 0 || sinks.Contains("github") {
 		// Cogito commit Github status mandatory fields.
+		if !src.GitHubApp.IsZero() && src.AccessToken != "" {
+			return fmt.Errorf("source: cannot specify both github_app and access_token")
+		}
+
+		// either access_token or github_app must be specified
+		if src.AccessToken == "" && src.GitHubApp.IsZero() {
+			return fmt.Errorf("source: one of access_token or github_app must be specified")
+		}
+
 		if src.Owner == "" {
 			mandatory = append(mandatory, "owner")
 		}
@@ -254,7 +263,7 @@ func (src *Source) Validate() error {
 			mandatory = append(mandatory, "repo")
 		}
 
-		if src.GitHubApp != (github.GitHubApp{}) {
+		if !src.GitHubApp.IsZero() {
 			if src.GitHubApp.ClientId == "" {
 				mandatory = append(mandatory, "github_app.client_id")
 			}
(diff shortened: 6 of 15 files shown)


aliculPix4D:
@scottillogical I would greatly appreciate if you can give us your feedback on the PR and also if you can give `pix4d/cogito:pci-4263-support-github-apps` a test run :) 

Thanks..

cogito/ghcommitsink.go
```diff
@@ -36,17 +36,27 @@ func (sink GitHubCommitStatusSink) Send() error {
 	ghState := ghAdaptState(sink.Request.Params.State)
 	buildURL := concourseBuildURL(sink.Request.Env)
 	context := ghMakeContext(sink.Request)
+	server := github.ApiRoot(sink.Request.Source.GhHostname)
+
+	token := sink.Request.Source.AccessToken
```
aliculPix4D, line 41: Note a fan of L41-L48... any feedback or improvement suggestion is greatly appreciated...

github/githubapp.go
```diff
@@ -0,0 +1,96 @@
...
+	PrivateKey     string `json:"private_key"` // SENSITIVE
+}
+
+// generateJWTtoken returns a signed JWT token used to authenticate as GitHub App
+func generateJWTtoken(clientId, privateKey string) (string, error) {
+	key, err := jwt.ParseRSAPrivateKeyFromPEM([]byte(privateKey))
+	if err != nil {
+		return "", fmt.Errorf("could not parse private key: %w", err)
+	}
+	// GitHub rejects expiry and issue timestamps that are not an integer,
```
aliculPix4D, line 27: typo: `expiry and issue timestamps`... It's not an `issue` timestamp... It	`Issued At` timestamp

cogito/protocol.go
```diff
@@ -245,7 +253,18 @@ func (src *Source) Validate() error {
 		if src.Repo == "" {
 			mandatory = append(mandatory, "repo")
 		}
-		if src.AccessToken == "" {
+
+		if src.GitHubApp != (github.GitHubApp{}) {
```
aliculPix4D, line 257: also not a fan of:

```
if ....
   if ...
else if ...
```
I tried many things... This is the cleanest so far... Any suggestion to improve?

cogito/protocol.go
```diff
@@ -162,9 +162,14 @@ type Source struct {
...
 	//
-	Owner       string `json:"owner"`
-	Repo        string `json:"repo"`
+	Owner string `json:"owner"`
+	Repo  string `json:"repo"`
+	// Mandatory if not using github app auth
 	AccessToken string `json:"access_token"` // SENSITIVE
+
+	// Mandatory if using github app auth
+	GitHubApp github.GitHubApp `json:"github_app,omitempty"`
```
aliculPix4D, line 171: if tried with pointers originally: `GitHubApp *github.GitHubApp `json:"github_app,omitempty"`` which made the following more natural:

 `if src.GitHubApp != nil ` but this brings other set of issues with panics due to nil pointer... Easily solved, but not worth it in my opinion...

cogito/protocol_test.go
```diff
@@ -238,6 +280,9 @@ repo:
 github_hostname:       
 access_token:          
 gchat_webhook:         
+github_app.client_id:        
+github_app.installation_id:  0
```
aliculPix4D, line 284: This 0 is the zero value for the int64 type... It really bothers my eyes :)

cogito/protocol.go
marco-m-pix4d: Note that `omitempty` and the newer `omitzero` are used on encoding, not decoding.

cogito/ghcommitsink_test.go
```diff
@@ -55,7 +113,7 @@ func TestSinkGitHubCommitStatusSendFailure(t *testing.T) {
 		Log:    testhelp.MakeTestLog(),
 		GitRef: "deadbeefdeadbeef",
 		Request: cogito.PutRequest{
-			Source: cogito.Source{GhHostname: gitHubSpyURL.Host},
+			Source: cogito.Source{GhHostname: gitHubSpyURL.Host, AccessToken: "dummy-token"},
```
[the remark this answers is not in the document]
aliculPix4D: Maybe we can do better here in the long run, by updating the test server handler to check that authorization token is set and set to expected value...

Otherwise:
```
			w.WriteHeader(http.StatusUnauthorized)
			fmt.Fprintln(w, "unauthorized: wrong token")
			return
```

Taskfile.yml
```diff
@@ -57,6 +57,10 @@ tasks:
       COGITO_TEST_REPO_OWNER: '{{default "pix4d" .COGITO_TEST_REPO_OWNER}}'
       COGITO_TEST_GCHAT_HOOK:
         sh: 'echo {{default "$(gopass show cogito/test_gchat_webhook)" .COGITO_TEST_GCHAT_HOOK}}'
+      COGITO_TEST_GH_APP_CLIENT_ID: '{{default "Iv23lir9pyQlqmweDPbz" .COGITO_TEST_GH_APP_CLIENT_ID}}'
+      COGITO_TEST_GH_APP_INSTALLATION_ID: '{{default "64650729" .COGITO_TEST_GH_APP_INSTALLATION_ID}}'
```
marco-m-pix4d, line 61: So the client ID would be always the same, representing `cogito` itself, and the installation ID would need to be created by everybody using the feature of GH app authentication ?

github/githubapp.go
```diff
@@ -0,0 +1,96 @@
...
+
+	jtwToken, err := generateJWTtoken(app.ClientId, app.PrivateKey)
+	if err != nil {
+		return "", err
+	}
+	req.Header.Set("Authorization", "Bearer "+jtwToken)
+
+	client := &http.Client{Timeout: time.Second * 5}
+
+	// FIXME: add retry here...
```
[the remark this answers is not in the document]
scottillogical: arent all github api requests rate limited?

cogito/ghcommitsink.go
marco-m-pix4d: Mhhh.
If the token is not set, then we do an HTTP request _each_ time to GitHub to generate this "installation token" ... Why not simply failing?

Taskfile.yml
scottillogical: are the defaults just for CI? I would not provide them at all

cogito/protocol.go
marco-m-pix4d: Now I also think I understand your question about the parenthesis around {}.

You can also add a method, say IsZero(), to this struct, and then modify the conditional to

```
if foo.IsZero() {
    ...
}
```

github/githubapp.go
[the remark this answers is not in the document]
aliculPix4D: Not these ones: `/app/installations/<ID>/access_tokens`. 

This was the very first thing I went to check... Otherwise, the whole GitHub app approach would not make sense in cogito...

With access token cogito is doing 1 POST request to update the status and we are limited to 5000 request per hour.

Imagine, with github app you get max 12500 request rate limit but now you must do 2 POST requests, 1 to fetch the installation token and 1 for the actual request you want to do...

If fetching installation token request counts towards the limit we would basically gain nothing...

cogito/protocol.go
marco-m-pix4d: I think that the current code is fine.

Note though I think a bug:
```
if src.GitHubApp != (github.GitHubApp{}) {
...
} else if src.AccessToken == "" {
...
}
```

it seems that src.AccessToken and src.GitHubApp  are mutually exclusive, but the code does not enforce this.

cogito/ghcommitsink.go
aliculPix4D: > Why not simply failing?

Then you would not use the github app token... 

Access token is the standard token that we were using until now.. If this value is empty that means that we are using github app, so we must generate installation token for github app..

cogito/protocol.go
marco-m-pix4d: I mean: I think we need something like (pseudocode):

if src.GitHubApp != (github.GitHubApp{}) && src.AccessToken != "" {
    error: cannot specify both
}

or something similar. But now I am even more confused, since we have
```
else if src.AccessToken == "" 
```
while I was expecting
```
else if src.AccessToken != "" 
```

Maybe I got it wrong...

github/githubapp.go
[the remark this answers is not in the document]
aliculPix4D: Note also:
https://docs.github.com/en/apps/creating-github-apps/about-creating-github-apps/best-practices-for-creating-a-github-app#cache-tokens

which shows that github apps are not really meant to be used for one-off tasks like in Concourse resources...

On each put step, we generate the installation token and throw it away at the end of the run... but token remains valid for 1 more hour. I hope there is no limit it number of access tokens an app can have...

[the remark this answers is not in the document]
aliculPix4D: In https://github.com/Pix4D/cogito/commit/10cdebae1f1edd1c918454d48b08bb23124566e8 I added the logging around both request... and then I triggered two subsequent builds..

Further info from experiments:

Logs build 1:

```
...
level=INFO msg=http-request name=cogito.put name=ghCommitStatus method=POST url=https://api.github.com/app/installations/64650729/access_tokens status=201 duration=173.116742ms rate-limit="" rate-limit-remaining="" rate-limit-reset=""
...
level=INFO msg=http-request name=cogito.put name=ghCommitStatus method=POST url=https://api.github.com/repos/pix4d/cogito-test-read-write/statuses/ea30d7fdcf39aa026bf54b4d2842692249317ab0 status=201 duration=268.61232ms rate-limit=10100 rate-limit-remaining=10096 rate-limit-reset=1745509359
```



Logs build 2:

```
...
level=INFO msg=http-request name=cogito.put name=ghCommitStatus method=POST url=https://api.github.com/app/installations/64650729/access_tokens status=201 duration=183.674616ms rate-limit="" rate-limit-remaining="" rate-limit-reset=""
...
level=INFO msg=http-request name=cogito.put name=ghCommitStatus method=POST url=https://api.github.com/repos/pix4d/cogito-test-read-write/statuses/ea30d7fdcf39aa026bf54b4d2842692249317ab0 status=201 duration=210.607387ms rate-limit=10100 rate-limit-remaining=10095 rate-limit-reset=1745509359
```

What we can see:
- the maximum rate limit for our organization test app is: `10100` which is correctly calculated according to GitHub docs, since we installed the app only on 1 test repo for now... Hence, only the number of users in our organization contribute towards increased rate limit. If we install the app on multiple repos the limit would further increase. In our case, reaching the maximum 12500 requests per hour
- in both builds we see that requests towards `https://api.github.com/app/installations/64650729/access_tokens` has no rate limit headers
- requests to `https://api.github.com/repos/pix4d/cogito-test-read-write/statuses/<sha>` consume the rate limit... 1 request per build as expected.

So everything works as expected. 
1. Global rate limit is increased from usual 5000 depending on how and where you install your app.
2. The request to generate the installation token is not rate limited.  
3. Rate limit is consumed only when you do the actual request that you wanted to do.
4. we do 2 requests instead of 1 so its normal that we will be a bit a slower...

cogito/ghcommitsink.go
aliculPix4D: > Why "we must generate installation token for github app.." each time that cogito runs ?

This is how github app works :)

You first use a private key and client id to generate a JWT token with very short validity...

https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation#using-an-installation-access-token-to-authenticate-as-an-app-installation

This enables you to authenticate as GitHub app and create a temporary installation token (valid for 1h; can't be changed) to authenticate as github app installation. 
https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation#authenticating-with-an-installation-access-token

Then you use this short-lived installation token to authenicate and do your API request that you wanted originally...

See also my comments in this thread below with logs:
https://github.com/Pix4D/cogito/pull/162#discussion_r2058506581

scottillogical:
I tested it and looks good, nice work

cogito/ghcommitsink.go
marco-m-pix4d: @aliculPix4D I read the documentation:

> The response will include an installation access token, the time that the token expires, the permissions that the token has, and the repositories that the token can access. The installation access token will expire after 1 hour.

So you are right, cogito must indeed do two API requests each time it runs, it is not possible to precompute the access token.

cogito/protocol.go
aliculPix4D: Fixed...

cogito/protocol.go
aliculPix4D: Fixed...

cogito/protocol.go
```diff
@@ -239,14 +247,33 @@ func (src *Source) Validate() error {
...
 		if src.Owner == "" {
 			mandatory = append(mandatory, "owner")
 		}
 		if src.Repo == "" {
 			mandatory = append(mandatory, "repo")
 		}
-		if src.AccessToken == "" {
-			mandatory = append(mandatory, "access_token")
+
+		if !src.GitHubApp.IsZero() {
```
aliculPix4D, line 267: so if we are here, either access_tolen or github_app is set; but we still need to check if all github_app mandatory fields are correctly configured if github_app is used...

Taskfile.yml
marco-m-pix4d: @aliculPix4D ?

cmd/cogito/main_test.go
```diff
@@ -134,6 +135,46 @@ func TestRunPutSuccessIntegration(t *testing.T) {
...
+	t.Setenv("ATC_EXTERNAL_URL", "https://cogito.invalid")
+	t.Setenv("BUILD_PIPELINE_NAME", "the-test-pipeline")
+	t.Setenv("BUILD_TEAM_NAME", "the-test-team")
+	t.Setenv("BUILD_NAME", "42")
+
+	err = mainErr(stdin, &stdout, &stderr, []string{"out", inputDir})
+
+	assert.NilError(t, err, "\nstdout:\n%s\nstderr:\n%s", stdout.String(), stderr.String())
+	assert.Assert(t, cmp.Contains(stderr.String(),
+		`level=INFO msg="commit status posted successfully" name=cogito.put name=ghCommitStatus state=error`))
```
marco-m-pix4d, line 175: Mhh.
Are we peeking in the logs like this in the other tests?

cogito/protocol.go
```diff
@@ -239,14 +247,33 @@ func (src *Source) Validate() error {
 	sinks := sets.From(src.Sinks...)
 	if sinks.Size() == 0 || sinks.Contains("github") {
 		// Cogito commit Github status mandatory fields.
+		// cannot specify both github_app and access_token
```
marco-m-pix4d, line 250: Comments at lines 250 and 255 seem to be repeating the error messages just two lines below them.

cogito/protocol_test.go
```diff
@@ -89,9 +104,45 @@ func TestSourceValidationFailure(t *testing.T) {
 
 	testCases := []testCase{
 		{
-			name:    "missing mandatory git source keys",
+			name:    "missing access token or github_app",
```
marco-m-pix4d, line 107: Ah. I got confused here. I think that a clearer test name is "access_token and github_app both missing".

github/githubapp.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/golang-jwt/jwt/v4"
+)
+
+type GitHubApp struct {
+	ClientId       string `json:"client_id"`
+	InstallationId int64  `json:"installation_id"`
+	PrivateKey     string `json:"private_key"` // SENSITIVE
+}
+
+func (app *GitHubApp) IsZero() bool {
```
marco-m-pix4d, line 21: Mhh. The logic in IsZero will return true if the struct is half constructed for any reason; this could hide a bug.
Was there a specific reason for this logic? Otherwise you could just compare with the zero value of the type, as it was in the previous version:

untested:
```
return *app == GitHubApp{}
```

github/githubapp.go
```diff
@@ -0,0 +1,109 @@
...
+// generateJWTtoken returns a signed JWT token used to authenticate as GitHub App
+func generateJWTtoken(clientId, privateKey string) (string, error) {
+	key, err := jwt.ParseRSAPrivateKeyFromPEM([]byte(privateKey))
+	if err != nil {
+		return "", fmt.Errorf("could not parse private key: %w", err)
+	}
+	// GitHub rejects expiry and issue timestamps that are not an integer,
+	// while the jwt-go library serializes to fractional timestamps.
+	// Truncate them before passing to jwt-go.
+	// Additionally, GitHub recommends setting this value 60 seconds in the past.
```
marco-m-pix4d, line 43: this value 60 -> this value to 60

github/githubapp.go
```diff
@@ -0,0 +1,109 @@
...
+	iat := time.Now().Add(-60 * time.Second).Truncate(time.Second)
+	// maximum validity 10 minutes. Here, we reduce it to 2 minutes.
```
marco-m-pix4d, line 45: - The maximum validity 10 minutes comes from where? Could we specify it?
- What is the meaning of `iat` ?

marco-m-pix4d: Ah, I see below. Then let's go against go and use readable names :-)

iat -> issuedAt
exp -> expiresAt

github/githubapp.go
```diff
@@ -0,0 +1,109 @@
...
+	exp := iat.Add(2 * time.Minute)
+	// Docs: https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app#about-json-web-tokens-jwts
+	claims := &jwt.RegisteredClaims{
+		IssuedAt:  jwt.NewNumericDate(iat),
+		ExpiresAt: jwt.NewNumericDate(exp),
+		// The client ID or application ID of your GitHub App.
+		// Use of the client ID is recommended.
```
marco-m-pix4d, line 52: Can go in the same line as 51

github/githubapp.go
```diff
@@ -0,0 +1,109 @@
...
+		return false
+	}
+	if app.PrivateKey != "" {
+		return false
+	}
+	return true
+}
+
+// generateJWTtoken returns a signed JWT token used to authenticate as GitHub App
+func generateJWTtoken(clientId, privateKey string) (string, error) {
```
marco-m-pix4d, line 35: if you pass the current time as a parameter, then this function is easily testable stand-alone.

github/githubapp.go
```diff
@@ -0,0 +1,109 @@
...
+	return token, nil
+}
+
+// GenerateInstallationToken returns an installation token used to authenticate as GitHub App installation
+func GenerateInstallationToken(server string, app GitHubApp) (string, error) {
+	// API: POST /app/installations/{installationId}/access_tokens
+	installationId := strconv.FormatInt(app.InstallationId, 10)
+	url := server + path.Join("/app/installations", installationId, "/access_tokens")
+
+	req, err := http.NewRequest(http.MethodPost, url, nil)
```
marco-m-pix4d, line 70: Ah. Don't we have a custom http request wrapper or convention? This I think will wait forever if something goes wrong.

github/githubapp.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	req.Header.Add("Accept", "application/vnd.github.v3+json")
+
+	jtwToken, err := generateJWTtoken(app.ClientId, app.PrivateKey)
+	if err != nil {
+		return "", err
+	}
+	req.Header.Set("Authorization", "Bearer "+jtwToken)
+
+	client := &http.Client{Timeout: time.Second * 5}
```
marco-m-pix4d, line 82: Here we care about the timeout, but we have line 70 before.
I think that we can share the client, it might even keep the TCP connection open.

aliculPix4D: Yes.. This is why I wrote the following comment in the original PR. See: https://github.com/Pix4D/cogito/pull/161/files#r2039756094

But, I think this would be a better fit to a separate PR... If you wish, I can first to a refactoring PR and sort this properly...

go.sum
```diff
@@ -11,6 +11,8 @@ github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/go-quicktest/qt v1.101.1-0.20240301121107-c6c8733fa1e6 h1:teYtXy9B7y5lHTp8V9KPxpYRAVA7dozigQcMiBust1s=
 github.com/go-quicktest/qt v1.101.1-0.20240301121107-c6c8733fa1e6/go.mod h1:p4lGIVX+8Wa6ZPNDvqcxq36XpUDLh42FLetFU7odllI=
+github.com/golang-jwt/jwt/v4 v4.5.2 h1:YtQM7lnr8iZ+j5q71MGKkNw9Mn7AjHM68uc9g5fXeUI=
```
aliculPix4D, line 14: for some reason it reverted this to v4... I wanted v5... I will update the PR...

github/githubapp.go
aliculPix4D: Note also that we have the same issues in github commit status 
https://github.com/Pix4D/cogito/blob/67e7fc9da35caef9c6c461eb265968734e23d880/github/commitstatus.go#L137

while in gchat sink we have request that can be terminated by context, but client with no timeout...
https://github.com/Pix4D/cogito/blob/67e7fc9da35caef9c6c461eb265968734e23d880/googlechat/googlechat.go#L99

cmd/cogito/main_test.go
aliculPix4D: In other integration tests... I copied and pasted it :) I think it was written by you actually :)

github/githubapp.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return "", fmt.Errorf("could not sign the JWT token: %w", err)
+	}
+	return token, nil
+}
+
+// GenerateInstallationToken returns an installation token used to authenticate as GitHub App installation
+func GenerateInstallationToken(server string, app GitHubApp) (string, error) {
+	// API: POST /app/installations/{installationId}/access_tokens
+	installationId := strconv.FormatInt(app.InstallationId, 10)
```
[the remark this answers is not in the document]
aliculPix4D: `fmt.Itoa` ??? you mean: `strconv.Itoa(app.InstallationId)`?

Anyway, I had `strconv.FormatInt` because `app.InstallationId` was int64... better question was why `app.InstallationId` was int64? Answer: no reason...

github/githubapp.go
aliculPix4D: Fixed... + added tests

github/githubapp.go
aliculPix4D: Fixed.

cogito/check_test.go
```diff
@@ -88,10 +89,25 @@ func TestCheckFailure(t *testing.T) {
 
 	testCases := []testCase{
 		{
-			name:    "validation failure: missing repo keys",
+			name:    "validation failure: missing access token or github_app",
 			source:  cogito.Source{},
 			writer:  io.Discard,
-			wantErr: "check: source: missing keys: owner, repo, access_token",
+			wantErr: "check: source: please specify either access_token or github_app",
```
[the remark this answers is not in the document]
aliculPix4D: Fixed.

cogito/protocol.go
aliculPix4D: Fixed.

github/githubapp.go
aliculPix4D: Fixed..

github/githubapp.go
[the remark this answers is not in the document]
aliculPix4D: Fixed by using int instead of in64

go.sum
aliculPix4D: Fixed...

github/githubapp.go
marco-m-pix4d: > while in gchat sink we have request that can be terminated by context, but client with no timeout...

the timeout in the http client struct has been added before "context", so if one is using "context", no need to set the timeout, and, only partially, vice-versa (only partially because "context" has effect on more that the timeout of the http client struct).

github/githubapp.go
aliculPix4D: Then I have to expose it to be able to test it...

Test package name: `package github_test`

github/githubapp.go
aliculPix4D: Yes but with timeout...

> client := &http.Client{Timeout: time.Second * 5}

github/githubapp.go
aliculPix4D: Ah this is not so simple... jwt package has some internal timers...

github/githubapp.go
aliculPix4D: See: https://github.com/Pix4D/cogito/pull/165

Taskfile.yml
aliculPix4D: Screenshot from 2025-04-29 09-56-34

Do you think this description is good enough? Otherwise, can you open: https://github.com/apps/pci-cogito-test and configure the description accordingly... I am not the GitHub organization admin... @odormond created the app for me... 

I do plan to update the contributing docs in this repo, mentioning the existence of this test app (next PR), but its still important that app description in GitHub UI is written correctly.

marco-m-pix4d: This is good enough, thanks.

github/githubapp.go
```diff
@@ -0,0 +1,99 @@
...
+func generateJWTtoken(clientId, privateKey string) (string, error) {
+	key, err := jwt.ParseRSAPrivateKeyFromPEM([]byte(privateKey))
+	if err != nil {
+		return "", fmt.Errorf("could not parse private key: %w", err)
+	}
+	// GitHub rejects expiresAt (exp) and issuedAt (iat) timestamps that are not an integer,
+	// while the jwt-go library serializes to fractional timestamps.
+	// Truncate them before passing to jwt-go.
+	// Additionally, GitHub recommends setting this value to 60 seconds in the past.
+	iat := time.Now().Add(-60 * time.Second).Truncate(time.Second)
```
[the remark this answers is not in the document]
aliculPix4D: Ah, it was not GitHub fault... I misunderstood you.. I thought you wanted me to correct this in comments only, not the variable names.

cogito/protocol.go
```diff
@@ -162,9 +162,14 @@ type Source struct {
...
 	//
-	Owner       string `json:"owner"`
-	Repo        string `json:"repo"`
+	Owner string `json:"owner"`
+	Repo  string `json:"repo"`
+	// Mandatory if not using github app auth
 	AccessToken string `json:"access_token"` // SENSITIVE
+
+	// Mandatory if using github app auth
+	GitHubApp github.GitHubApp `json:"github_app"`
```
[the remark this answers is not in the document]
aliculPix4D: As you already saw, yes the private_key is a sensitive value and we redact it below with:

`fmt.Fprintf(&bld, "github_app.private_key:      %s\n", redact(src.GitHubApp.PrivateKey))`

github/githubapp.go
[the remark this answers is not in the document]
aliculPix4D: Fixed...

github/githubapp.go
```diff
@@ -0,0 +1,99 @@
...
+	// Github set the maximum validity of a token to 10 minutes. Here, we reduce it to 1 minute
+	// (we set expiresAt to 2 minutes, but we start 1 minute in past).
```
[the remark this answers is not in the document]
aliculPix4D: Fixed..

marco-m-pix4d (approved):

aliculPix4D:
@odormond @iAmoric @SpectreVert still missing review on this one... Please have a look.

cmd/cogito/main_test.go
```diff
@@ -134,6 +135,46 @@ func TestRunPutSuccessIntegration(t *testing.T) {
...
+		},
+		Params: cogito.PutParams{State: cogito.StateError},
+	}))
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+	inputDir := testhelp.MakeGitRepoFromTestdata(t, "../../cogito/testdata/one-repo/a-repo",
+		testhelp.HttpsRemote(github.GhDefaultHostname, gitHubCfg.Owner, gitHubCfg.Repo), gitHubCfg.SHA,
+		"ref: refs/heads/a-branch-FIXME")
+	t.Setenv("BUILD_JOB_NAME", "TestRunPutGhAppSuccessIntegration")
+	t.Setenv("ATC_EXTERNAL_URL", "https://cogito.invalid")
```
odormond, line 166: I don't think the value actually matters but did you chose the `invalid` TLD on purpose?
I'm asking because it is a reserved one with special behaviour defined by [RFC6761 section 6.4](https://www.rfc-editor.org/rfc/rfc6761.html#section-6.4). If there is no real intent to have an NXDOMAIN response, maybe using `example` would be better (see [RFC6761 section 6.5](https://www.rfc-editor.org/rfc/rfc6761.html#section-6.5).

github/githubapp.go
```diff
@@ -0,0 +1,99 @@
...
+func (app *GitHubApp) IsZero() bool {
+	return *app == GitHubApp{}
+}
+
+// generateJWTtoken returns a signed JWT token used to authenticate as GitHub App
+func generateJWTtoken(clientId, privateKey string) (string, error) {
+	key, err := jwt.ParseRSAPrivateKeyFromPEM([]byte(privateKey))
+	if err != nil {
+		return "", fmt.Errorf("could not parse private key: %w", err)
+	}
```
odormond, line 31: Why are we parsing it that late instead of while validating the config?

odormond (approved):

cmd/cogito/main_test.go
aliculPix4D: I must say: I don't have any smart answer except... I copy/pasted it from existing tests...

https://github.com/Pix4D/cogito/blob/fa53eabfaba7dff0d6de9fc94306e3c7c5f81e57/cmd/cogito/main_test.go#L123

But, you are correct, I remember the same discussion in some other project (skyguide maybe?)... Anyway... I will not address this comment in this PR, but I can prepare a separate PR to fix it everywhere...

@odormond ok?

odormond: Fine for me. Definitely a very minor point.

github/githubapp.go
aliculPix4D: @odormond do you mind if I keep it as it is for this PR, and try to propose another separate PR addressing this in a more focused way?

odormond: Fine for me.

PR #166, opened 2025-04-30
update docs and acceptance tests pipeline

aliculPix4D:
update docs and acceptance tests pipeline

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,9 +183,38 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
 - `repo`\
   The GitHub repository name.
 
+###  Authentication keys
+
+Keys must be provided for one of the following authentication methods:
+1. [Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)
+2. [GitHub app](https://docs.github.com/en/apps/creating-github-apps/about-creating-github-apps/about-creating-github-apps)
+
+#### Personal Access Token keys
+
 - `access_token`\
   The OAuth access token.\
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -376,6 +399,14 @@ NOTE: The token is security-sensitive. Treat it as you would treat a password. D
 
 See also the section [Integration tests](./CONTRIBUTING.md#integration-tests) for how to securely store the token to run the end-to-end tests.
 
+# GitHub App auth
+
+1. Create a new github app  https://github.com/settings/apps/new
```
marco-m-pix4d, line 404: github app -> GitHub App

README.md
```diff
@@ -183,10 +183,33 @@ With reference to the [GitHub Commit status API], the `POST` parameters (`state`
 - `repo`\
   The GitHub repository name.
 
+###  Authentication keys
+
+Keys must be provided for one of the following authentication methods:
```
marco-m-pix4d, line 188: If I am a new user and I arrive here, I don't know which one to pick.
In this section, just after the list with the 2 options, we could say something to help the user choose.

Maybe something along these lines:

```
Which authentication method to choose?

The advantage of using a Personal Access Token is that it is simpler to setup,
while the advantage of a GitHub App is that it has a more generous rate limit.
See section [#Caveat: GitHub rate limiting] for details.
```

marco-m-pix4d (approved):

README.md
aliculPix4D: Fixed as suggested. Thanks

README.md
aliculPix4D: Fixed...

odormond (approved):

iAmoric (approved):

PR #168, opened 2025-05-02
replace .invalid domain name with .example

aliculPix4D:
- replace `.invalid` domain name with `.example`

Addresses the comment in previous PR: https://github.com/Pix4D/cogito/pull/162#discussion_r2069896024


odormond (approved):
:bow:

PR #177, opened 2026-04-08
googlechat: retry on temporary failures and on timeouts

marco-m-pix4d:
googlechat: retry on temporary failures and on timeouts

The API of googlechat.TextMessage is now a bit clumsy.
I plan another PR to simplify; on the other hand, this should be ready to be deployed.

Best reviewed commit-per-commit.

PCI-4096

### DIFF
--- a/googlechat/googlechat.go
+++ b/googlechat/googlechat.go
@@ -14,8 +14,11 @@ import (
 	"log/slog"
 	"net/http"
 	"net/url"
+	"slices"
 	"strings"
 	"time"
+
+	"github.com/Pix4D/go-kit/retry"
 )
 
 // BasicMessage is the request for a Google Chat basic message.
@@ -57,8 +60,22 @@ type MessageSpace struct {
 	DisplayName string `json:"displayName"` // Name of the space in the UI.
 }
 
+// DefaultRetry returns a [retry.Retry] with the recommended values to be passed to
+// [TextMessage] for production. If you have special requirements, or for testing,
+// you can override completely or partially.
+func DefaultRetry(log *slog.Logger) retry.Retry {
+	upTo := 60 * time.Second
+	return retry.Retry{
+		UpTo:         upTo,
+		FirstDelay:   1 * time.Second,
+		BackoffLimit: upTo / 4,
+		Log:          log,
+	}
+}
+
 // TextMessage sends the one-off message `text` with `threadKey` to webhook `theURL` and
 // returns an abridged response.
+// Use [DefaultRetry] for parameter 'rtr'.
 //
 // Note that the Google Chat API encodes the secret in the webhook itself.
 //
@@ -74,6 +91,7 @@ type MessageSpace struct {
 func TextMessage(
 	ctx context.Context,
 	log *slog.Logger,
+	rtr retry.Retry,
 	webHook, threadKey, text string,
 ) (MessageReply, error) {
 	body, err := json.Marshal(BasicMessage{Text: text})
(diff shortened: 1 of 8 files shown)


googlechat/googlechat_test.go
```diff
@@ -30,15 +31,107 @@ func TestTextMessageIntegration(t *testing.T) {
...
+		assert.ErrorContains(t, err, http.StatusText(code))
+		assert.Equal(t, sleepTimeSpy, wantSlept)
+	}
+
+	test(http.StatusForbidden, 0)                 // not retriable: fails immediately.
+	test(http.StatusTooManyRequests, rtr.UpTo)    // retriable; fails after consuming all retries.
+	test(http.StatusServiceUnavailable, rtr.UpTo) // retriable; fails after consuming all retries.
+}
+
+// FIIIIIIXME For rubustness needs test time bubble!!!!
```
marco-m-pix4d: Delete this line

googlechat/googlechat.go
```diff
@@ -133,6 +140,57 @@ func TextMessage(
 	return reply, nil
 }
 
+var retryables = []int{
+	http.StatusTooManyRequests,     // 429
+	http.StatusInternalServerError, // 500
+	http.StatusServiceUnavailable,  // 503
+	http.StatusGatewayTimeout,      // 504
+}
```
marco-m-pix4d, line 154: Maybe also 

		http.StatusBadGateway,          // 502

???

marco-m-pix4d: For example, curl:

> Transient error means either: a timeout, an FTP 4xx response code or an
>              HTTP 408, 429, 500, 502, 503 or 504 response code.

408 Request Timeout
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/408

aliculPix4D (approved):
Nice...

googlechat/googlechat.go
odormond: For this it's probably fine to retry on more errors. Worst case, we post the message multiple times.

In general, I think retrying on 502 is safer than on 500 or 504. A prior, 502 means the gateway is not configured properly so retrying might hit the gateway after it got fixed but 500 and 504 are usually not safe for requests that change the server state as we don't know if the faulty request managed to alter it or not and repeating the alteration unconditionally might have consequences.

googlechat/googlechat.go
```diff
@@ -133,6 +140,57 @@ func TextMessage(
...
+			bytes.NewBuffer(body))
+		if err != nil {
+			return fmt.Errorf("TextMessage: new request: %w", RedactErrorURL(err))
+		}
+		req.Header.Set("Content-Type", "application/json; charset=UTF-8")
+		// Encode the thread Key as a URL parameter.
+		if threadKey != "" {
+			values := req.URL.Query()
+			values.Set("threadKey", threadKey)
+			req.URL.RawQuery = values.Encode()
```
odormond, line 175: Do we really need to recompute the URL on each retry? AFAIU, we always produce the same URL.

odormond (approved):

googlechat/googlechat.go
marco-m-pix4d: Eh.
I arrived here because before I was getting unexplainable errors from the tests. Me too I don't like it.
I am adding a note to try to get something less ugly in the follow-up work.

googlechat/googlechat.go
marco-m-pix4d: I understand the concern here, especially around 504, but how come curl considers all these codes safe ?
(see the extract from curl man page just above, found for `curl --retry`. Curl does have a "retry everything", under `--retry-all-errors`) ...

It seems every author has their notion of what is safe. For example urllib3:
```
RETRY_AFTER_STATUS_CODES = frozenset({413, 429, 503})
```

Update: given the intended usage, I am keeping the list by curl, but disabling 504.

PR #180, opened 2026-04-13
ci: finally cache Go dependencies; try ratchet to pin GH Actions

marco-m-pix4d:
1. ci: finally cache Go dependencies
It took me years of experiments to finally find the (obvious in retrospective) fix to the warning

    `Restore cache failed: Dependencies file is not found in /home/runner/work/cogito/cogito. Supported file pattern: go.mod`

This seems to have reduced the CI run from `1m30s` to `1m`.

2. try ratchet to pin GH Actions

Done with:

    ratchet pin .github/workflows/ci.yml

ratchet: https://github.com/sethvargo/ratchet

PCI-4096

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # ratchet:actions/checkout@v6
         with:
           # By default, actions/checkout will persist the GITHUB_TOKEN, so that further
           # steps in the job can perform authenticated git commands (that is: WRITE to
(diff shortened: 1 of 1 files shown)


odormond (approved):

.github/workflows/ci.yml
```diff
@@ -28,6 +22,14 @@ jobs:
           # the repo). Following the Principle of least privilege, we disable this as long
           # as we don't need it.
           persist-credentials: false
+      # Installing Go must be done AFTER checking out the code, in order to have
```
iAmoric, line 25: Could you explain a bit more why/how it works?

iAmoric (approved):

aliculPix4D (approved):

SpectreVert (approved):

.github/workflows/ci.yml
marco-m-pix4d: From the PR description, I was always getting the following warning from the Go Action:

```
Restore cache failed: Dependencies file is not found in
/home/runner/work/cogito/cogito. Supported file pattern: go.mod
```

and I kept banging my head against the wall: File `go.mod` is there, why are you complaining ??? I tried N variations of the settings available to the Go Action, but no change.

At the end I realized that when the Go Action was running, since the repo was not there yet, it indeed could not find the `go.mod` file.

marco-m-pix4d: A classic 🤦

iAmoric: Ah, as simple as that...

aliculPix4D: > At the end I realized that when the Go Action was running, since the repo was not there yet, it indeed could not find the go.mod file.

This is worrying to me... I would expect that actions/setup-go action just installs and configures Go. Why the hell would it require the source code?

